PR #3, opened 2016-09-14
Updated minified and added project-specific annotation file

sgosline:
We reviewed these in our meeting today.


sgosline:
Also added in patient/sample identifier...

PR #16, opened 2016-10-14
Added in a yml to create file manifest from repo

sgosline:
Focused on NF-specific work.


sgosline:
Also updated annotation files to move out project-specific items from minified.

PR #22, opened 2016-10-18
Update minimal_Sage_standard.json

kdaily:
Add `mirnaSeq` value.


kkdang:
Group approves addition of term.

PR #23, opened 2016-10-20
Added Not Applicable and Control to minified

sgosline:
Updated json and then created yml to match. Also updated ncats annotations.


ntap_annotations/ntap_ncats_annotation_scheme.yml
```diff
@@ -41,8 +41,8 @@ entityId1: syn5611797, syn5611796
 filePath1: '10x10 Combination Screens/HFF'  ##this is an alternate way of referencing an entity if localOrRemote is remote, than file path refers to head of synapse project.
      annotations:
```
kkdang: Sara, do you want these data structures to be a list containing dictionary items? Alternatively, you can have a dictionary of dictionaries if you remove the hyphen.

sgosline: No, let's just leave it as it is because Xindi has already built a script to create a manifest around it.

PR #39, opened 2017-01-10
Script to convert JSON to Synapse Table Schema

kdaily:


kdaily:
See https://www.synapse.org/#!Synapse:syn7444884/tables/ for imported examples.

PR #41, opened 2017-01-24
Add GENIE annotations

thomasyu888:


genie_annotations.json
```diff
@@ -0,0 +1,24 @@
+{
+  "tissueSource": [
```
kdaily: see tissue and organ in minified standard.

genie_annotations.json
```diff
@@ -0,0 +1,24 @@
...
+    "NKI",
+    "DFCI",
+    "GRCC",
+    "JHU",
+    "MSK",
+    "UHN",
+    "VICC",
+    "MDA"
+  ],
+  "platform": [
```
kdaily: I believe we discussed platform being controlled in the minified standard? If not we should.

kkdang: Platform is a key in the project-level dictionaries; values are thus not controlled (currently). We might not want it in minified, since platforms could vary widely across projects using different techologies.

genie_annotations.json
kkdang: We'll need to remove sample-specific annotations from this dictionary for other reasons anyway.

genie_annotations.json
kdaily: True, but enabling discovery of data on the same platform across projects is important. We should draw from a standardized source of platform names; `HiSeq2000` should be, at a minimum, `IlluminaHiSeq2000`, for example (https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL11154)

kkdang: HiSeq2000 is consistent with platform names in the neurannotations dictionary that we have already accepted. We can discuss standardization of these names at a future meeting, but I think for now it is okay to accept since it is consistent across projects.

kdaily: Move platform to minified, please!

PR #63, opened 2017-03-07
Create ngs_annotations.json

amapeters:
NGS assay annotations


amapeters:
I thought we wanted to split the assays up? Leave assay in minified, and assay specific terms in separate schema. What about assay specific platforms?

sgosline:
The assay-specific _keys_ should be split up - the term 'assay' should only be defined in one file, and I believe this is still a minified.

sgosline:
Also, can you please move the platform to this file too? I feel like that's ngs specific, no?

amapeters:
Yes, but then we end up with the platform key in multiple files

sgosline:
Aside from confocalImaging (which can arguably be put somewhere else) everything is either array or sequencing - perhaps we can call it arrayPlatform and sequencePlatform and put them in the respective tech-specific dictionaries?

amapeters:
we have already stared to use the key 'platform' - can edit if all agree we should, but we need to lock this down soon

ngs_annotations.json
```diff
@@ -0,0 +1,45 @@
+{
+    "assay": [
```
kdaily: assay stays in minified

ngs_annotations.json
```diff
@@ -0,0 +1,45 @@
...
+        false 
+       
+    ],
+    "libraryPrep": [
+        "rRNAdepletion",
+        "polyAselection",
+        "lncRNAenrichment",
+        "miRNAenrichment"
+    ],
+    "fileSubFormat": [
```
kdaily: goes to neuro (STAR aligner specific)

ngs_annotations.json
```diff
@@ -0,0 +1,45 @@
...
+        "rRNAdepletion",
+        "polyAselection",
+        "lncRNAenrichment",
+        "miRNAenrichment"
+    ],
+    "fileSubFormat": [ 
+        "TranscriptomeSAM",
+        "SortedByCoordinate"
+    ],
+    "libraryPrep": [
```
kdaily: delete this one

kdaily:
Review me please, @kkdang!

kdaily:
Or @sgosline!

Issue #71, opened 2017-03-21
Organize repo


sgosline:
We should organize the repos

- create directory structure for binaries and schema
- make sure there are jsons and remove ymls
- Update readme to describe how to use these schemas.

jaeddy:
I'm inclined to organize the repo following standard structure for a Python package. This might be overkill, as we might not ever need to import or work with annotation "objects" within Python, but this would leave us with the option to move in that direction.

The most relevant example is the GA4GH schemas repo/library:
https://github.com/ga4gh/ga4gh-schemas/tree/master/python

The purpose and use cases of their schemas are definitely more advanced and ambitious than ours will (likely) be; however, the idea of defining schemas in terms of Python classes would have some advantages in terms of testing, inheritance, and extensibility.

Here's my proposal for the initial re-org, with some comments below:

```
synapseAnnotations/
├── LICENSE
├── README.md
├── bin
├── docs
│   └── use_cases.yml
├── scripts
│   ├── createTermList.py
│   └── json2schema.py
├── synapseAnnotations
│   ├── __init__.py
│   └── data
│       ├── Neuro_Annotations.json
│       ├── genie_annotations.json
│       ├── minimal_Sage_analysis.yml
│       ├── minimal_Sage_standard.json
│       ├── minimal_Sage_standard.yml
│       ├── nf_annnotations.yml
│       └── nf_annotations.json
└── tests
```

- `synapseAnnotations` is where the main library code would go; because this code doesn't really exist yet, and the schemas currently only exist as files, I've placed them in a `data` subfolder and dropped in an empty `__init__.py` file. The library itself will likely comprise core functionality for working with schemas and schema files and (eventually, maybe) code/modules that actually define the schema (replacing the current JSON/YAML versions).  
- `bin` is where production-level wrapper scripts would live; these import the `synapseAnnotations` library (or individual modules) and perform common tasks/operations.
- `scripts` is a place for in-development code which might eventually become modules in the main library or production-level scripts in `bin`
- `docs` is obvious; I think because there will be a lot of use cases we want to document and demonstrate, building out a set of pages with Sphinx and ReadTheDocs could be nice.
- `tests` a friendly reminder that we should have them, in some form, sooner than later!

Not currently shown/included: other Python packaging elements (e.g., `setup.py`, etc.).

I think that's it for now. Please comment below.

teslajoy:
^ very nice!

kdaily:
[...]

PR #86, opened 2017-03-30
Restructure repo as Python package

jaeddy:
Creates a main library folder (synapseAnnotations) to store schema-
related Python modules and data (JSON/YAML files). Production-ready
scripts go in 'bin' and in-progress scripts in 'scripts'. Also adds
a 'docs' folder (for more content later). Still need to add tests
and 'setup.py'.

Resolves: #71


jaeddy:
I wasn't sure how much hierarchy was desired &mdash; everything gets validated against common, or each level gets validated against its parent level. For example, if a project can have multiple assays, then do we have nested folders? If, on the other hand, assay-related JSONs don't depend on project-related JSONs (or vice versa), then the org that @kdaily suggested seems good.

kdaily:
I believe the latter is what was discussed (separate assay and project - only one layer of folders to separate). Will pose this ? in the slack to confirm. Thanks!

sgosline:
I like the _idea_ of the directory hierarchy, but perhaps we can delay that until we have a more concrete strategy for key-conflict checking.

PR #87, opened 2017-03-30
Add tests

kdaily:
Using `nose` to test.

First test checks for duplicate values within each key - there were some problems with Neuro, removed duplicates to get test to pass.

Could use feedback on introspectively determining where the JSON files to test are.


jaeddy:
This is a philosophical side note, not a review: I would argue that the `test_duplicate_values_within_key()` might actually belong in some sort of schema validation module (i.e., part of the package codebase). You'd then have a corresponding unit test to make sure that the validation function performs as expected (e.g., given some toy/mock examples with or without dups). I still think we should move ahead with this for now, but something to keep in mind.

Also, something something (PEP8)[https://www.python.org/dev/peps/pep-0008/#maximum-line-length]... ;)

kdaily:
Noted, @jaeddy - will convert as we discuss the rest of things to test, after I convert all of the JSON files to the new schema format. Thanks!

kdaily:
That only works to tell you which key has duplicate values, but not which are the values duplicated. Thanks!

jaeddy:
👍  Ah right.. paying closer attention to your exception message would have been smart.

PR #93, opened 2017-04-03
list to json - what happened there @-@?

teslajoy:


teslajoy:
Lets kill the request for now. and/or add to a TODO/issue.  This format has a different use-case than what I needed (combining multiple annotations into one dictionary). That requires call to .update() and for all .json files to be in a dictionary format... not a list of objects :) woot!

PR #103, opened 2017-04-04
Add fileFormat descriptions

jaeddy:
Adds descriptions and URLs to values under the `fileFormat` key.


synapseAnnotations/data/common/minimal_Sage_standard.json
```diff
@@ -138,1156 +138,1156 @@
...
+    "name": "individualID",
+    "description": "",
+    "columnType": "STRING",
+    "maximumSize": 250,
     "enumValues": []
-  }, 
+  },
   {
-    "name": "fileFormat", 
-    "description": "",
```
sgosline: Shouldn't you actually describe the key itself?

PR #111, opened 2017-04-06
Add tool annotations

jaeddy:
Not ready to merge yet, but opening a PR to make it easier for us to discuss and iterate.

cc: @sgosline, @kdaily, @blogsdon


sgosline:
My first use case is to annotate the scripts we have uploaded already:
1- User wants to query for tools of a certain type of data (e.g. image data) in a certain language (matlab).  
2- user wants to compare all tools across to analyze geneExpression data

jaeddy:
Adding to @sgosline's use cases, I'm thinking of two others at the moment:

1. If new tools are developed as part of a project/consortium, we'd want to report and annotate them in a similar way to newly generated data (this connects to Sara's example, in that including annotation creates more value and utility for the shared tools, rather than just highlighting them).

2. Rather than a replacement for provenance, what I had in mind was more of an "augmented" provenance. The `used` in an activity would point to a Synapse entity, which would either point to an external file (e.g., GitHub) or something stored directly in Synapse. The annotations then provide a way to summarize/categorize what the script/tool was designed to do as well as details of how it was created, instead of requiring curious users to dig through the code to figure it out themselves. I expect to have an example of this latter use case for AMP-AD within the next ~week.

synapseAnnotations/data/tool_annotations.json
```diff
@@ -0,0 +1,46 @@
...
+        "Shell"
+    ],
+    "scriptLanguageVersion": [],
+    "softwareRepository": [
+        "Synapse",
+        "GitHub",
+        "GitLab",
+        "BitBucket"
+    ],
+    "softwareRepositoryType": [
```
sgosline: What does this mean @jaeddy?

sgosline:
I like the idea of separating out the extended - can we make the terms mutually exclusive?  I will probably only use the core terms.

Issue #112, opened 2017-04-07
Should we remove annotations that are always empty?


sgosline:
In the update/sync command, can we update the schema to remove all keys for which the value is always blank? Talked to @xindiguo about this, wanting more feedback. I realize this belongs in the synAnnotationsUtils project, but I'm secretly hoping to move those scripts to this repo ;)

kdaily:
:confused:

kkdang:
Sara, can you move this issue to the other repo until (if) they become merged?

sgosline:
Are we going to formally review issues on that repo? It looks like it hasn't been touched in weeks.

PR #115, opened 2017-04-11
additional cell types and assay types for NF2 datasets

allaway:
I added a couple of assay types and cell types that are used frequently in the Synodos NF2 data. I also added an "unknown" to NF1/NF2 genotype in NF_annotations as the data do not always exist to confirm genotype status

### DIFF
--- a/synapseAnnotations/data/common/minimal_Sage_standard.json
+++ b/synapseAnnotations/data/common/minimal_Sage_standard.json
@@ -138,13 +138,13 @@
         "source": "https://www.illumina.com/techniques/sequencing/dna-sequencing/targeted-resequencing/targeted-panels.html"
       },
       {
-        "value": “jumpingLibrary”, 
+        "value": "jumpingLibrary", 
         "description": "", 
         "source": ""
       },
       {
-        "value": “MIB/MS”, 
-        "description": “Kinomics assay using multiplexed kinase inhibitor beads and mass spectrometry (MIB/MS)”, 
+        "value": "MIB/MS", 
+        "description": "Kinomics assay using multiplexed kinase inhibitor beads and mass spectrometry (MIB/MS)", 
         "source": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3328787/"
       }
     ]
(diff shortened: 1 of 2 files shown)


synapseAnnotations/data/common/minimal_Sage_standard.json
```diff
@@ -1045,6 +1055,41 @@
         "value": "CD8+ T-Cells",
         "description": "",
         "source": ""
+      },
+      {
+        "value": "CRISPR arachnoid",
```
sgosline: Is this arachnoid with a CRISPR mutation of a gene?  If so, should probably be arachnoid and use the ONC annotations to apply the crispr info, no?

synapseAnnotations/data/common/minimal_Sage_standard.json
```diff
@@ -1045,6 +1055,41 @@
...
+        "description": "",
+        "source": ""
+      },
+      {
+        "value": "primary arachnoid",
```
sgosline: Is this to distinguish between cultured? If so both cells are arachnoid and isCellLine is altered...

synapseAnnotations/data/common/minimal_Sage_standard.json
allaway: Yes- i think that's the best solution. I had added these before I looked at your recent pull request.

synapseAnnotations/data/common/minimal_Sage_standard.json
allaway: primary is still cultured, but (in most cases) low passage, heterogenous cell mixtures. eg. disaggregated tumor cells being grown in culture, rather than a more homogenous cell "line"

synapseAnnotations/data/common/minimal_Sage_standard.json
allaway: So should these be also rolled into the NF_annotations or minified?

synapseAnnotations/data/common/minimal_Sage_standard.json
sgosline: Got it, ok.

synapseAnnotations/data/common/minimal_Sage_standard.json
sgosline: I think NF should supplement onc, and then when we apply annotations we just ignore all keys for which there are no values (see issue #112 )

allaway:
@kdaily all up to date, could you merge? i have no write access

PR #130, opened 2017-04-19
added isPrimary key and few other values to minified

allaway:
This is to clear up the issue of having primary cells in cellTypes. There are four cells I added in my last pull request that are primary or the not primary equivalent (arachnoid, primary arachnoid, meningioma, primary meningioma). 
@teslajoy removed the primary arachnoid and primary meningioma for me, and I'm adding an isPrimary boolean to make it easy to flag these cell lines (allowing me to just annotate these lines as arachnoid or meningioma, with isPrimary True or False).

I also added a couple other changes that I mentioned in issues in the past week.


allaway:
I can't merge this PR, just FYI

PR #149, opened 2017-06-12
added UConn Stem Cell Core value for specimenIdSource key

ychae:
for Broad iPSC


sgosline:
Too detailed, closing.

PR #151, opened 2017-06-13
added transcriptAssemblyMethod key and cufflinks value

ychae:


kdaily:
Should go into `analysis.json` and be called `readAssemblyMethod` per meeting today. @amapeters @ychae

Issue #155, opened 2017-06-20
Revisit minified to determine what is needed


sgosline:
Perhaps the minified is not as minimal as we like - we should revisit.

sgosline:
Use case examples:
1- User wants to find all RNA-Seq datasets that study colon cancer tumors, in fastq form
2- User wants to compare glioma tumor samples (WGS/exome) to germline samples from patients with neurodegenerative disorders
3- User wants to compare their own data to all NF1 samples in Synapse

sgosline:
See pull request #161  - I reduced minified to support a more hierarchical scheme, somewhat inspired by the organization by GA4GH here: https://github.com/ga4gh/ga4gh-schemas/tree/master/src/main/proto/ga4gh 

Most existing minified went into experimental annotations.

sgosline:
Minified is no more with #161 - now sageCommunities...

PR #161, opened 2017-07-02
Proposal for reduced minified

sgosline:
This is a possible solution for #155 - created new minified term called 'resource type' that will allow for different types of resources.  Most minified (with exception of tumorType) got moved to experimental annotations.  Tumor type got moved to onc.


sgosline:
Ok, updated file name to avoid minified claim.  Also moved files out of common.

kdaily:
@blogsdon 

thanks

PR #165, opened 2017-07-17
Added ability for RESTful API

sgosline:
Some people want their RESTful API to be linked on Synapse. This change reflects the annotations.


kdaily:
Think we chatted about this one - that RESTs can only be APIs?

kdaily:
@sgosline think this existed before the big move happened. Please either fix or close, and re-open as necessary. Thanks!

sgosline:
Closing until we have a specific use case.

PR #167, opened 2017-07-25
Added fileFormat - idat

amapeters:


kdaily:
@amapeters got source and changed the description to match the text from that source.

PR #170, opened 2017-07-27
Added - 'channel' red and green

amapeters:


sgosline:
I'd argue that this does not belong in an 'ngs' module, but rather an array-based module of annotations so that it is not required of all ngs datasets...

kdaily:
Agreed. I will add an array file and make this the first term.

kdaily:
@amapeters I changed to `Cy3` and `Cy5` as these are the actual channels/labels, and have existing term sources available.

PR #195, opened 2017-08-28
Added individualIDSource values

amapeters:


synapseAnnotations/data/experimentalData.json
```diff
@@ -309,6 +309,35 @@
     "columnType": "STRING",
     "maximumSize": 250,
     "enumValues": []
```
sgosline: Need to leave enum open, and remove duplicates.

PR #197, opened 2017-08-31
added values narrowPeak, broadPeak, and gappedPeak to fileFormat key

ychae:


kdaily:
See this example as 'bed' as a prefix in ENCODE data matrix:

https://www.encodeproject.org/matrix/?type=Experiment&status=released&award.project=ENCODE&files.file_type=bed+narrowPeak

kdaily:
@amapeters is that OK?

kdaily:
@kkdang the prefixes didn't make it in, though? Might want to revert and restore this PR before?

kkdang:
rats, yes, I need to do that

PR #199, opened 2017-09-04
changed most http to https sources 

teslajoy:
but can not do much about links with http web servers on hosts  end. 

I did go through all available annotations.


teslajoy:
add curl -I url test to docs 
only change synapse links to be https

PR #202, opened 2017-09-06
Added cellType - oligodendrocytes

amapeters:


kdaily:
The term in the ontology is singular, so changed that - also changed the URL to the original source instead of the EBI lookup (should change the rest of them that have that pattern too)

PR #212, opened 2017-09-15
Added fileFormat - sh

amapeters:


kdaily:
`sh` should be for any shell script, not just bash. I'm looking for a better source.

kdaily:
I suppose Wikipedia is *slightly* better than the current source:

https://en.wikipedia.org/wiki/Shell_script

I noted above that `.sh` is not just for bash scripts, but for other shells as well. You cannot determine the language of the shell script by file extension, only by examining the shebang (first line). Similarly, as the Wikipedia page notes, many/most shell scripts don't actually have a file extension, they are just set to be executable.

If you want to make it clear that it really is a bash shell script (which is a good idea, but starts to get into the realm of tools (as @sgosline mentioned at the last meeting), then the value for this should be `bash script` instead of `sh`.

PR #229, opened 2017-09-27
renamed key "nf1/nf2genotype"

allaway:
"genotype" doesn't capture all of the genetic perturbations in the NF1/NF2 datasets. for example, HS01 cells have constitutively expressed shRNA targeting NF2 that results in extreme NF2 deficiency, but it is still NF2 +/+ 

renaming the key expands it's utility. could explore changing "-/-" to "deficient"in the scenario illustrated above.

### DIFF
--- a/synapseAnnotations/data/neurofibromatosis.json
+++ b/synapseAnnotations/data/neurofibromatosis.json
@@ -41,22 +41,22 @@
       {
         "value": "-/-", 
         "description": "Homozygous deletion, mutation or other form of NF2 deficiency (eg shRNA knockdown of NF2)", 
-        "source": ""
+        "source": "Sage Bionetworks"
       }, 
       {
         "value": "+/-", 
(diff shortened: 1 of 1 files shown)


allaway:
have discussed this with @sgosline

synapseAnnotations/data/neurofibromatosis.json
```diff
@@ -33,14 +33,14 @@
...
+    "name": "nf2Status", 
     "description": "Genotype of NF2 gene, if known", 
     "columnType": "STRING", 
     "maximumSize": 250, 
     "enumValues": [
       {
         "value": "-/-", 
-        "description": "Homozygous deletion or mutation", 
+        "description": "Homozygous deletion, mutation or other form of NF2 deficiency (eg shRNA knockdown of NF2)", 
         "source": ""
```
kkdang: Please include source: "Sage Bionetworks"

PR #238, opened 2017-10-11
Added analysisType - DNAmethylationImputation

amapeters:
need description and source


kdaily:
@amapeters did you find a source and description?

amapeters:
No, can we reference a publication?

kdaily:
I have no problem with that - use a DOI url? Would prefer a second opinion before confirming.

Issue #242, opened 2017-10-17
CHiP-seq is spelled wrong


kdaily:

kdaily:
Closed by #243.

PR #243, opened 2017-10-17
correcting ChIPseq spelling 

allaway:
#242


allaway:
@kdaily it looks like travis failed here. Not sure why?

kdaily:
Because I force merged before it finished, and the reference went away in the middle of travis doing stuff.

kdaily:
Note to self: don't force merge so these things don't happen again.

patience

PR #251, opened 2017-10-18
Added cellType 'neuron'

amapeters:


kdaily:
reopen against develop @amapeters k thx bye

PR #265, opened 2017-10-24
Added tissue - 'precuneus cortex'

amapeters:


kdaily:
In issue #259 @blogsdon requested to use the term from NCIT, not UBERON. @amapeters can you confirm with him please?

kdaily:
Should be 'precuneus'. See #259 please.

PR #271, opened 2017-10-27
New drug terms

sgosline:
Working to move things around to manage drug screening data.


synapseAnnotations/data/compoundScreen.json
```diff
@@ -0,0 +1,72 @@
+[
+  {
+    "name": "compoundName",
```
sgosline: @allaway Can you please add the drug names/sources you provided for me in json format?

allaway: @sgosline yes - how do you want me to resolve the cases where the source compound name is different from the name used in the data?

PR #281, opened 2017-10-31
analysisType - enrichmentAnalysis

amapeters:


sgosline:
shouldn't this match the link to be 'Enrichment analysis'? Or are we sticking with camel case despite the standards?

kdaily:
@sgosline correct, fixed!

Issue #233, opened 2017-10-10
SOP for annotating cell lines by individualId and sampleId


sgosline:
It will be helpful if we have a standard convention for annotating cell line data. For example - are cell lines individuals? or samples? We have some cell lines that are derived from the same patient but that's not necessarily the status quo. 

I'd like to establish a convention so we can stick to it.

kdaily:
In PCBC and OHSU-LINCS we had standalone annotations for cell line. Important to note OHSU-LINCS annotates down to the aliquot (including growth conditions), but that is not propagated to Synapse.

I feel it's important to put into context of data use - if patient info is important, then cell line may have to go to `sampleId`.

sgosline:
There is also a cellLine annotation for the PSON cell line repository. It's worth discussing adding it to the set, but what do we do for sampleId and patientId for cell line data?

sgosline:
Decision is to create new terms "cellLine" and "cellLineSource" to experimentalData.json.

PR #291, opened 2017-11-28
addition of cell Line term

sgosline:
Addition of cell line name to address issue #233


sgosline:
Any update on this changelog issue? Should we just close and create a new one?

sgosline:
@kdaily ?

CHANGELOG.md
```diff
@@ -496,4 +496,4 @@
 
 
 
-\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
```
kdaily: @karawoo could you please help figure out why this changed? it's the only reason we haven't merged this - wanted to avoid making changes to the changelog outside of the release process.

karawoo: I have no idea, but the content is not actually changed at all, so I'm not too worried about it. I can rebase this branch to just keep the change to `experimentalData.json` if we want to prevent it from looking as though the changelog has been changed.

PR #315, opened 2018-01-16
Updated analysis annotations

sgosline:
Added new annotations to describe the ongoing NF-related analysis that will be used to build dashboards.


sgosline:
@allaway will remove regression and add elastic net to supervised learning method.

Issue #305, opened 2017-12-15
compoundDose values


allaway:
```
  {
      "name": "compoundDose",
      "description": "",
      "columnType": "DOUBLE",
      "enumValues": []
  },
```

I have a use case where I am annotating a file with a single compound but a range of log(M) compound doses (-4 to -7.3 log(M)). 

I'd like to annotate the range. Should this be a new key (compoundDoseRange) or should we modify the above key? I'd annotate the compoundDoseUnit as log(M). 

@sgosline

sgosline:
I'm not sure what to do here. I think for dose response curves we need the range, but for images and so forth we only need a single concentration. I like the idea of adding a second key, as much as it pains me.

allaway:
Agreed; will put in PR for new key.

PR #322, opened 2018-01-30
add compoundDoseRange

allaway:
Adding compoundDoseRange, a string that represents the min-max values of a drug/compound/antibody treatment range.


allaway:
closes #305

PR #346, opened 2018-02-25
Added missing dictionary items to key

teslajoy:
Key has been skipped from neuro module in annotations table v8.0.0 due to missing maxSize and enumVal list items.


teslajoy:
Should we unify the structure of our dictionary objects here for booleans type?

kdaily:
@teslajoy absolutely!

kdaily:
It means downstream usage might change - we will manually have to recommend the values to use (`true` and `false`).

PR #355, opened 2018-03-02
Revise data, experimentalData and curatedData

allaway:
Created data.json and curatedData.json
moved some terms from experimentalData to data as these should apply to both curated and experimental data.

Added some terms to curatedData that would be useful for my use-cases and some others that I am aware of.

added curatedData as a resourceType in sageCommunity.

in thinking about this more, clinicalData might be worth having in it's own file...


allaway:
@kdaily this should fulfill your requested changes

PR #358, opened 2018-03-07
added ngs analysis terms

allaway:
transcriptQuantificationMethod : bedtools multicov
differentialExpressionMethod : edgeR


allaway:
@sgosline is this in reference to both values or just bedtools multicov?

allaway:
@sgosline woo!

PR #370, opened 2018-04-05
Update CONTRIBUTING.md

kdaily:
add note about which term to use from source - the name as given, or a synonym.


kdaily:
Oops, thanks @allaway! Merging.

PR #373, opened 2018-04-06
Added tissue value cortical plate

ychae:


ychae:
@kdaily, the source has been updated :)

Issue #351, opened 2018-02-28
Think more and come to some conclusion about JSON Schema and JSON-LD


kdaily:
Placeholder for now. Briefly discussed ways of/reasons for modularizing the annotations. A common solution is JSON-LD:

https://json-ld.org/

jaeddy:
+1

sgosline:
🤔

sgosline:
...
🤔 
...
🤔 
...

sgosline:
Done.

jaeddy:
The Human Cell Atlas (HCA) project seems to use JSON-LD for their schemas (see [comment](https://github.com/Sage-Bionetworks/synapseAnnotations/issues/339#issuecomment-369983263) in #339)). I don't quite get how it works yet, but might be a useful example for us to check out.

sgosline:
@jaeddy and @kdaily and @karawoo will test drive some of our annotations using json-LD to get a feel for how it works.

sgosline:
Oh wait, now you're asking for conclusions? Jeeezzzz

sgosline:
Ooh, a cartoon description: https://www.youtube.com/watch?v=4x_xzT5eF5Q

sgosline:
Would this work as a validator? https://github.com/Julian/jsonschema??

sgosline:
Another proposal from @jaeddy: store ontologies in OWL format (https://github.com/stain/owlapi-jsonld) - places like http://www.obofoundry.org/ store all ontologies in this format so we can go between custom/general formats with same tools.

karawoo:
I edited the title of this issue because I think so far we've really been talking more about using JSON Schema than using JSON-LD, and as far as I can tell HCA isn't using JSON-LD in any of their examples. @jaeddy I'd definitely love to talk about how we want to store our own ontologies. Maybe it's time for another meeting?

jaeddy:
I didn't read this comment carefully until this morning. You're totally right! I don't think that JSON Schema and JSON-LD are mutually exclusive at all, but it seems like JSON Schema is a better fit for specifying (and validating) our... schema. 

JSON-LD could enhance the functionality of annotation data, as JSON objects/documents generated based on our JSON Schema structure would include context and references. Whether this is something implement for managing annotations or ask the engineering team to support in Synapse is probably worth "thinking about" in a separate issue.

[...]

PR #375, opened 2018-04-24
Beginning framework of json LD

sgosline:
Just to link to #351 for a point of reference


jaeddy:
Need to figure out how to handle the whole ontology thing, but made some initial progress on NGS annotations.

karawoo:
I move we close this PR since it's so old. I think with our newer thinking about JSON Schema and JSON-LD we'd need to mostly start over from scratch anyway. Any objections @sgosline?

sgosline:
nope.

PR #387, opened 2018-05-23
added NHP-Chimpanzee study to neuro

ychae:


ychae:
@amapeters can you review and approve this PR?

PR #395, opened 2018-06-15
Updated terms for JMML spore data

### LINKED ISSUE #394: Update annotations to account for JMML DHART SPORE project
From Elliot:
```
For disease, I would include juvenile myelomonocytic leukemia.

For organ (and tissue), buccal mucosa cell and marrow fibroblast should be options.

For cell type, is should probably be bone marrow and buccal mucosa cell as well….
```
Need to figure out what works!

sgosline:
closes #394

### DIFF
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -311,7 +311,7 @@
       {
         "value": "skin",
         "description": " An organ that constitutes the external surface of the body. It consists of the epidermis, dermis, and skin appendages.",
-        "source":"http://purl.obolibrary.org/obo/NCIT_C12470",
+        "source":"http://purl.obolibrary.org/obo/NCIT_C12470"
       }
     ]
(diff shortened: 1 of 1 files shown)


synapseAnnotations/data/experimentalData.json
```diff
@@ -307,6 +307,11 @@
         "value": "bone marrow",
         "description": "The soft, fatty, vascular tissue that fills most bone cavities and is the source of red blood cells and many white blood cells.",
         "source": "http://purl.obolibrary.org/obo/BTO_0000141"
+      },
+      {
+        "value": "skin",
```
karawoo: We already have skin here: https://github.com/Sage-Bionetworks/synapseAnnotations/blob/534c82b81d33562ea8d6bfda9cc1c3b2dfa621b8/synapseAnnotations/data/experimentalData.json#L251-L255

Is that definition not sufficient?

PR #396, opened 2018-06-27
NF-related additions

sgosline:
A few updates to account for NF data

### DIFF
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -215,6 +215,64 @@
     ]
   },
   {
+    "name": "assayTarget", 
+    "description": "", 
+    "columnType": "STRING", 
+    "maximumSize": 250, 
+    "enumValues": [
+      {
+        "value": "H3K9me3", 
+        "description": "", 
+        "source": ""
+      }, 
+      {
+        "value": "H3K4me3", 
+        "description": "", 
+        "source": ""
+      }, 
+      {
+        "value": "H3K4me1", 
+        "description": "", 
+        "source": ""
+      }, 
+      {
+        "value": "H3K27ac", 
+        "description": "", 
+        "source": ""
+      }, 
+      {
+        "value": "H3K36me3", 
+        "description": "", 
+        "source": ""
+      }, 
+      {
+        "value": "H3K27me3", 
+        "description": "", 
+        "source": ""
+      }, 
+      {
+        "value": "H3K9ac", 
+        "description": "", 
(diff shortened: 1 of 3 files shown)


sgosline:
updated to move assayTarget

karawoo:
@sgosline is Dorsal Root Ganglion really a cell type, or a cluster of cells? Most of the other entries in `cellType` look like types of individual cells, but I'm not really sure...

sgosline:
That's a good point, the full cells are "DRG/nerve root neurosphere cells"

sgosline:
So maybe DRG is the tissue...

karawoo:
That makes more sense to me but it might be good to get confirmation from someone else. Everything else in this PR looks good to me so if you want to pull out that one change for now I will approve it ✅

sgosline:
Hmm, I'll tag @allaway. Robert, the description of the cells is here: https://www.synapse.org/#!Synapse:syn8016635/wiki/411638

allaway:
DRG is structured and comprised of  multiple cell types, so I think `tissue` is appropriate for it (although NCIT considers it a `part` of the PNS). 

This paper from Lu's lab describes the model in more detail. https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4254535/

As I understand it, DRG/nerve root neurosphere cells are still probably comprised of multiple cell types/subtypes. Since it's only the embryonic GAP43+PLP+ Schwann cell precursors that are being isolated and expanded, perhaps `embryonic GAP43+PLP+ Schwann cell precursor` is the most descriptive cell type. `Schwann cell precursor` could also be sufficient - that's what this group calls GAP43+PLP+ SC precursor cells: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5469943/.

Cell Ontology has that as a listed term: http://purl.obolibrary.org/obo/CL_0002375

synapseAnnotations/data/experimentalData.json
```diff
@@ -937,6 +1005,11 @@
     "columnType": "STRING",
     "maximumSize": 250,
     "enumValues": [
+      {
+        "value": "Dorsal Root Ganglion",
```
allaway: Move to "tissue"

synapseAnnotations/data/experimentalData.json
```diff
@@ -937,6 +1005,11 @@
     "columnType": "STRING",
```
allaway: add `Schwann cell precursor`

http://purl.obolibrary.org/obo/CL_0002375

PR #400, opened 2018-07-12
Added value hic to fileFormat

ychae:


ychae:
This should be good to go, it just needs an approving review.

ychae:
Thanks @karawoo !

PR #407, opened 2018-08-13
Add whole brain tissue

karawoo:
Still a little torn on whether it's better to use "whole brain" (more specific) vs. just "brain" (which is an existing term we can reuse)...

### DIFF
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -800,6 +800,11 @@
         "value": "Dorsal Root Ganglion",
         "description": "Ganglion with sensory function within the vertebral column.",
         "source": "http://purl.obolibrary.org/obo/NCIT_C12462"
+      },
+      {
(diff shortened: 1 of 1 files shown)


karawoo:
Sorry I'm confused. You _do_ want it to be "brain" instead of "whole brain", or...?

karawoo:
Discussed in person with @sgosline:

- `"brain"` alone is not ideal because brain is already an organ and we prefer to not have organ and tissue be the same (except for blood)
- another option would be to have an option like `"unspecified"` which could be used for any sample that comes from an organ but isn't from one specific tissue. A downside is that on its own it's not very meaningful.

karawoo:
Outcome from meeting: we will add `tissue:unspecified` for any tissue from any organ where the specific tissue is not specified. 

We will probably also add `organ:unspecified` so that we can capture whole body parts (i.e. whole head)

PR #408, opened 2018-08-13
Additions to entries in species, assay, platform, organ in experimentalData.json; additions to tumorType in cancer.json 

milen-sage:
Perhaps we need a more generic term for methylation sequencing instead of scCGIseq (the latter is a new method that has not been added to more standard ontologies, yet)? For now I added it under assays, but we can always substitute it.

### DIFF
--- a/synapseAnnotations/data/cancer.json
+++ b/synapseAnnotations/data/cancer.json
@@ -235,7 +235,7 @@
       {
         "value": "Lymphoblastic Lymphoma",
         "description":"A lymphoma composed of immature small to medium-sized precursor lymphoid cells (lymphoblasts). It includes the B- and T-cell lymphoblastic lymphoma.",
-        "source": "https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&version=18.06d&ns=ncit&code=C9360&key=n4324201&b=1&n=null"
+        "source": "http://purl.obolibrary.org/obo/NCIT_C9360"
       },
       {
         "value": "Meningioma",
--- a/synapseAnnotations/data/cancer.json
+++ b/synapseAnnotations/data/cancer.json
@@ -233,6 +233,11 @@
         "source": "https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0000693"
       },
       {
+        "value": "Lymphoblastic Lymphoma",
+        "description":"A lymphoma composed of immature small to medium-sized precursor lymphoid cells (lymphoblasts). It includes the B- and T-cell lymphoblastic lymphoma.",
+        "source": "https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&version=18.06d&ns=ncit&code=C9360&key=n4324201&b=1&n=null"
+      },
+      {
         "value": "Meningioma",
         "description": "A generally slow growing tumor attached to the dura mater. It is composed of neoplastic meningothelial (arachnoidal) cells.",
         "source": "https://www.ebi.ac.uk/ols/ontologies/mp/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C3230"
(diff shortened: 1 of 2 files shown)


synapseAnnotations/data/experimentalData.json
```diff
@@ -171,11 +171,16 @@
...
         "source": ""
       },
-      {
+       {
         "value": "MIB/MS",
         "description": "Kinomics assay using multiplexed kinase inhibitor beads and mass spectrometry (MIB/MS)",
         "source": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3328787/"
       },
+      {
+        "value": "scCGIseq",
```
allaway: I think this is fine. See above (MIB/MS) for example of specific sequencing method. I think with assays the more specific, the better (not an official contributing requirement AFAIK, though).

synapseAnnotations/data/experimentalData.json
```diff
@@ -435,6 +445,11 @@
         "source": ""
       },
       {
```
sgosline: Agree with @allaway here, epigenetics is so vague I wasn't allowed to use it in a manuscript title (that talked about two various forms of epigenetic activity) - please stick to chromatin activity.

milen-sage: Agreed :)

synapseAnnotations/data/experimentalData.json
```diff
@@ -941,6 +941,11 @@
         "description": "",
         "source": ""
       },
+      {
+        "value": "AffymetrixHuman1.0STS",
```
allaway: I wasn't able to find an Affymetrix Human 1.0 STS mentioned anywhere with a google search. Are you sure this is not Affymetrix Human ST? 

GEO is a great source for platform sources: https://www.ncbi.nlm.nih.gov/geo/browse/?view=platforms

synapseAnnotations/data/cancer.json
```diff
@@ -232,6 +232,11 @@
         "description": "A neoplasm that arises from SCHWANN CELLS of the cranial, peripheral, and autonomic nerves. ",
         "source": "https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0000693"
       },
+      {
+        "value": "Lymphoblastic Lymphoma",
+        "description":"A lymphoma composed of immature small to medium-sized precursor lymphoid cells (lymphoblasts). It includes the B- and T-cell lymphoblastic lymphoma.",
+        "source": "https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&version=18.06d&ns=ncit&code=C9360&key=n4324201&b=1&n=null"
```
allaway: Would suggest instead using the OBO link for source: http://purl.obolibrary.org/obo/NCIT_C9360

synapseAnnotations/data/experimentalData.json
```diff
@@ -211,6 +216,11 @@
         "value": "Ribo-Seq",
         "description": "Ribosome profiling (Ribo-Seq).",
         "source": "http://www.ebi.ac.uk/efo/EFO_0008891"
+      },
+      {
+        "value": "RPPA",
```
allaway: would suggest the short link: http://www.bioassayontology.org/bao#BAO_0010030

synapseAnnotations/data/experimentalData.json
```diff
@@ -375,6 +385,11 @@
         "value": "bone marrow",
         "description": "The soft, fatty, vascular tissue that fills most bone cavities and is the source of red blood cells and many white blood cells.",
         "source": "http://purl.obolibrary.org/obo/BTO_0000141"
+      },
+      {
+        "value": "Bursa Of Fabricius",
+        "description": "An epithelial outgrowth of lymphoid tissue in the cloaca of young birds, responsible for B-lymphocyte maturation. It undergoes atrophy and involution at six months of age.",
+        "source": "https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&version=18.06d&ns=ncit&code=C111141&key=n1891607686&b=1&n=null"
```
allaway: same comment as above - would recommend short link

synapseAnnotations/data/experimentalData.json
```diff
@@ -1361,6 +1381,11 @@
         "value": "Rhesus macaque",
         "description": "Macaca mulatta with taxonomy ID: 9544 and Genbank common name: Rhesus monkey",
         "source": "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=9544&lvl=3&p=has_linkout&p=blast_url&p=genome_blast&p=mapview&lin=f&keep=1&srchmode=1&unlock"
+      },
+      {
+        "value": "Chicken",
```
allaway: Would instead suggest gallus gallus. A preferred source for these is probably the NLM taxonomy browser (given the other entries).

allaway:
looks good!

PR #411, opened 2018-08-14
Update sageCommunity.json

kdaily:
we use 'bgzip', so we should use 'gzip' here and not use the common extension, but a better description of what the format is!

### DIFF
--- a/synapseAnnotations/data/sageCommunity.json
+++ b/synapseAnnotations/data/sageCommunity.json
@@ -431,9 +431,9 @@
         "source": "http://edamontology.org/format_2332"
       },
       {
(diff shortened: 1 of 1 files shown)


ychae:
The source link http://www.ebi.ac.uk/swo/data/SWO_3000077 seems to be borken. Is it working for you @kdaily ?

karawoo:
Doesn't work for me

kdaily:
Bah, I notified EBI in January that all SWO links are broken. Seems like they haven't fixed it.

karawoo:
Closing this as the changes were manually added to #410

Issue #405, opened 2018-08-07
Add embryo under a new semantic type?


milen-sage:
Currently we don't have embryonic structure as a semantic type in experimental data annotations, e.g. 
https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&version=18.06d&ns=ncit&code=C28147&key=1009529467&b=1&n=null

A few studies use embryonic stem cells/embryo tissues, and we might need corresponding annotation entries?

allaway:
Is this resolved?

milen-sage:
Yes, that got resolved with the addition of embryonic tissue to 'tissue' and adding various embryonic cell types to 'cellType'.

PR #413, opened 2018-08-16
Added new keys to assay, platform, tissue, organ, tumorType; updated ngs.json to fix downstream annotation upload (via annotator)

milen-sage:


synapseAnnotations/data/experimentalData.json
```diff
@@ -750,6 +755,11 @@
         "description": "",
         "source": ""
       },
+      {
+        "value": "embryonic",
```
allaway: this probably merits further discussion - is this a tissue? or a feature of a tissue? for example, what about embryonic blood? or embryonic brain?

karawoo: Agree, we should definitely discuss

milen-sage: I added a related issue #405 to our next milestone! :)

karawoo: Agreed in 2018-08-28 meeting that this should be renamed to "embryonic tissue"

milen-sage: Updated to embryonic tissue.

synapseAnnotations/data/cancer.json
```diff
@@ -197,6 +197,11 @@
         "description": "A malignant, usually aggressive tumor composed of atypical, neoplastic melanocytes.",
         "source": "https://www.ebi.ac.uk/ols/ontologies/mp/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C3224"
       },
+      {
+        "value": "Glioma",
```
karawoo: I wish there were some way to nest `"Low Grade Glioma"` and `"Glioblastoma"` under `"Glioma"` 😕 but as there isn't this term is ok with me.

synapseAnnotations/data/cancer.json
```diff
@@ -247,6 +252,11 @@
         "description": "Acute lymphoblastic leukemia of T-cell origin. It comprises about 15% of childhood cases and 25% of adult cases. ",
         "source":"http://purl.obolibrary.org/obo/NCIT_C3183"
       },
+      {
+        "value": "Chronic Lymphocytic Leukemia",
```
karawoo: Could you match the capitalization from http://purl.obolibrary.org/obo/DOID_1040?

synapseAnnotations/data/cancer.json
milen-sage: Yes, agreed.

allaway: A possible solution to this would be to start using oncotree codes (where low grade glioma would be LGGNOS, glioma would be GNOS, and glioblastoma multiforme would be GBM 
(where NOS means "not otherwise specified").
more here:
http://oncotree.mskcc.org/#/home

allaway:
Looks good to me.

PR #418, opened 2018-09-13
add datatype weight

allaway:
This repo is like an astronaut - weightless. 

We have `assay = scale` but no equivalent `dataType`.


allaway:
woo!

PR #420, opened 2018-09-18
Update neuro.json

kelshmo:
Removed group key because no longer in use.


karawoo:
@kdaily @amapeters can you confirm that we're not using this key across any neuro projects? I don't think I am using it.

kdaily:
We use it in BSMN. The groups involved have not been added to this key, though.

karawoo:
Maybe we should keep the key and add a description but remove the enumerated values and just let it be any string?

kdaily:
@karawoo I like the proposal!

kdaily:
See #422.

kelshmo:
No objections! This initial request to remove the key back in September came out of a conversation with Yooree.

kdaily:
OK! I'm going to close this and I'll make a new request that does what is proposed above. Thanks @kelshmo @karawoo!

PR #422, opened 2018-09-21
add file formats

allaway:
zip, mov, hyperlink (for external links to google sheets, docs, etc manifested as a a synapse entity), svs, ab1, 7z, gff3, gson, sqlite, svg.

### DIFF
--- a/synapseAnnotations/data/sageCommunity.json
+++ b/synapseAnnotations/data/sageCommunity.json
@@ -476,14 +476,14 @@
         "source": "https://www.loc.gov/preservation/digital/formats/fdd/fdd000470.shtml"
       },
       {
-        "value": "ab1",
+        "value": "AB1",
         "description": "TAB1 binary format of raw DNA sequence reads (output of Applied Biosystems' sequencing analysis software). Contains an electropherogram and the DNA base sequence.",
         "source": "http://edamontology.org/format_3000"
       },
       {
-        "value": "7z",
-        "description": "7z is a compressed archive file format that supports several different data compression, encryption and pre-processing algorithms.",
-        "source": "https://en.wikipedia.org/wiki/7z"
+        "value": "7Z Compression Format",
+        "description": "A compressed archive file format that supports several different data compression, encryption and pre-processing filters.",
+        "source": "http://purl.obolibrary.org/obo/NCIT_C80224"
       },
       {
         "value": "gff3",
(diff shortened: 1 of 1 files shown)


synapseAnnotations/data/sageCommunity.json
```diff
@@ -454,6 +474,36 @@
...
+      },
+      {
+        "value": "ab1",
+        "description": "TAB1 binary format of raw DNA sequence reads (output of Applied Biosystems' sequencing analysis software). Contains an electropherogram and the DNA base sequence.",
+        "source": "http://edamontology.org/format_3000"
+      },
+      {
+        "value": "7z",
+        "description": "7z is a compressed archive file format that supports several different data compression, encryption and pre-processing algorithms.",
+        "source": "https://en.wikipedia.org/wiki/7z"
```
karawoo: Maybe http://purl.obolibrary.org/obo/NCIT_C80224 instead?

synapseAnnotations/data/sageCommunity.json
```diff
@@ -454,6 +474,36 @@
...
+      },
+      {
+        "value": "json",
+        "description": "JavaScript Object Notation format; a lightweight, text-based format to represent tree-structured data using key-value pairs.",
+        "source": "http://edamontology.org/format_3464"
+      },
+      {
+        "value": "sqlite",
+        "description": "SQLite formatted file.",
+        "source": "Sage Bionetworks"
```
karawoo: http://edamontology.org/format_3621

synapseAnnotations/data/sageCommunity.json
allaway: Yeah...I saw this one but "7Z Compression Format" seems so excessive!

I'll make the change, though

allaway:
@karawoo re SVS I opted for "svs" because it looks like other microscope software (non-Aperio) can also use this format: v

To make matters even more confusing, Aperio doesn't really exist anymore as an entity- they were acquired by Leica back in 2012.

synapseAnnotations/data/sageCommunity.json
karawoo: It's ok with me to keep the value as "7Z" even though the preferred term is "7Z Compression Format". This is what we did with "gzip", since we're already inconsistent on this with other fileFormat annotations (like tsv) and many of our fileFormat values don't come from ontologies at all.

allaway:
OK - that sounds good to me! I'll change it back.

allaway:
woohoo!

PR #436, opened 2018-12-19
Add some AMP-AD related annotation values

karawoo:
* Add `"Resilience"` as a consortium (@amapeters should this be `"Resilience-AD"`?)
* Add `"JAX"` as an individualIdSource


karawoo:
@amapeters should be good now

PR #437, opened 2018-12-19
Add ELISA as assay and immunoassay as dataType

karawoo:
I've used "immunoassay" (an exact synonym) rather than "Immunology Test" (the preferred term) because it appears that "immunoassay" is already in use in AMP-AD. However if we'd prefer to switch to the preferred term I'm happy to update.


sgosline:
Interesting, I would have thought immunoassay was too specific, but it checks out. Go for it.

PR #438, opened 2018-12-19
Add behavioral assays

karawoo:
("contextual conditioning behavior" has no description on the term page)

### DIFF
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -222,7 +222,31 @@
         "source": "http://purl.obolibrary.org/obo/NBO_0000217"
       },
       {
-
+        "value": "contextual conditioning behavior",
+        "description": "",
+        "source": "http://purl.obolibrary.org/obo/NBO_0000223"
+      },
+      {
+        "value": "anxiety-related behavior",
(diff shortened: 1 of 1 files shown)


sgosline:
@milen-sage do any of these correspond with the piggy tests?

sgosline:
otherwise it looks good from the NTAP perspective.

karawoo:
No she didn't

milen-sage:
> @milen-sage do any of these correspond with the piggy tests?
The anxiety-related behavior is related to the piggy tests. And @sgosline you are right that overall that would be a good category to accommodate the pig-characteristics tests for composite pain scores. I can add the piggy assays to this category (they are not really pig specific in terms of keys).

amapeters:
There is a definition in Ontobee, which is the source we have used for the other behavioral assays: http://www.ontobee.org/ontology/NBO?iri=http://purl.obolibrary.org/obo/NBO_0000223

karawoo:
? that is the source I used, and I don't see a description on that page

karawoo:
Summarizing in-person conversation with @amapeters: only contextual conditioning behavior is missing a description, the others each have one, and we should try to keep using this neuro behavior ontology for consistency with previous behavioral assays. Based on this I think this PR should be approved unless anyone else has specific concerns.

Issue #120, opened 2017-04-12
Add annotations for reference genome build/assembly


jaeddy:
For any processed data, it'd be great to have some tags indicating genome build (e.g., hg19 vs. GRCh38), possibly as fine-grained as version (e.g., patch release 10 vs. patch release 9). I suggest adding a key to minified based on the GA4GH class [`ReferenceSet`](http://ga4gh-schemas.readthedocs.io/en/latest/schemas/references.proto.html#protobuf.ReferenceSet).

For example:

```
[
  ...
  {
    "name": "referenceSet",
    "description": "",
    "columnType": "STRING",
    "maximumSize": 250,
    "enumValues": [
      {
        "value": "GRCh38", 
        "description": "Genome Reference Consortium Human Build 38", 
        "source": "https://www.ncbi.nlm.nih.gov/assembly/GCF_000001405.26"
      },
  ...
]
```

Accounting for version could get complicated, as you might have to deal with patch number as well as contributing source (e.g., GRC vs. UCSC). Gets even trickier if you include reference _transcriptomes_ (i.e., FASTA sequences derived from gene models in Ensembl, RefSeq, GENCODE, etc. that all have their own release versions as well). Would need to give this some more thought...

kdaily:
@jaeddy if this is still required please open a PR with changes to `ngs.json`, thanks!

sgosline:
Let's discuss this in the context of scRNA-Seq annotations

allaway:
This is related to the issue we discussed today, #350

sgosline:
It will go into curatedData.json for now.

kelshmo:
Do we have consensus of how to annotate with this key? I believe it makes sense to add values for both Ensembl and UCSC ref genomes.

karawoo:
I believe this was closed in #439

PR #439, opened 2018-12-19
Add referenceSet to curatedData.json

kelshmo:
Annotation pulled from issue #120.


karawoo:
@jaeddy any chance you could take a quick look at this and see if it makes sense?

synapseAnnotations/data/curatedData.json
```diff
@@ -38,5 +38,18 @@
         "source": "http://semanticscience.org/resource/CHEMINF_000123"
       }
     ]
+  },
+  {
+    "name": "referenceSet",
+    "description": "",
```
jaeddy: I would suggest this description (paraphrased from old GA4GH [schemas](https://github.com/ga4gh/ga4gh-schemas/blob/master/src/main/proto/ga4gh/references.proto)):

> "A set of references (e.g., canonical assembled contigs) which defines a common coordinate space for comparing reference-aligned experimental data."

jaeddy:
I assume that this annotation would only be relevant when `"curatedDataType"` = `"reference sequence"`?

Issue #17, opened 2016-10-14
Code for automated checking of keys between minimal and project-specific files


kdaily:
Discussed that the existing annotations code could probably solve this - @xindiguo?

sgosline:
This would go away if we go with the idea proposed in #134

kdaily:
Implement in travis - should block a merge.

sgosline:
Add to contributing.md.

PR #440, opened 2019-01-10
add initial version of VEO-IBD specific annotation.

kdaily:
We are using 'group' in the same way that AMP-AD is right now. Added values for `group`.

### DIFF
--- a/synapseAnnotations/data/veo-ibd.json
+++ b/synapseAnnotations/data/veo-ibd.json
@@ -1,0 +1,65 @@
+[
+  {
+    "name": "group", 
+    "description": "Groups involved in the VEO-IBD consortium.", 
+    "columnType": "STRING", 
+    "maximumSize": 250, 
+    "enumValues": [
+      {
+        "value": "Hospital for Sick Children", 
+        "description": "", 
+        "source": "Sage Bionetworks"
+      }, 
+      {
+        "value": "Dr. von Hauner Children’s Hospital", 
+        "description": "", 
+        "Source": "Sage Bionetworks"
+      }, 
+      {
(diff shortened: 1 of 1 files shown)


kdaily:
Fair point - that should have been 'neuro' (as defined in `neuro.json`) not AMP-AD. But VEO-IBD is not neuro, hence the standalone. There are multiple groups from the same institution; but some of those groups are also technically groups (with a different subset of individuals) that are in other neuro projects!

allaway:
Ah, yeah. I guess what I was getting at is that if a key description is shared across communities or json, shouldn't the description be agnostic to the json it is in? like this definition instead, without the cancer specificity: https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C48168

 i.e. `description: A collection of people with a common research objective.`

I realize this is super nitpicky and probably not necessary, but figure I'd bring it up just in case. :)

karawoo:
I think it is important @allaway, because otherwise we could end up with the same key defined in two different, incompatible ways, with no way to determine which one was intended.

kdaily:
Not nitpicky at all! We've discussed and punted on this too many times to count, though :P

@karawoo the initial proposed solution was an integration test that wouldn't allow duplicate values for the same key across JSON files, but that has yet to be implemented. Then, first one in gets it, and latter people have to change/obfuscate/etc if they collide.

kdaily:
Indeed, #17 suggests this (only open by me for > 2 years).

karawoo:
@kdaily do you know of other keys that are currently already duplicated across the json files?

PR #442, opened 2019-01-22
unset enum values for group

kdaily:
Proposal instead of #420 which removed entire key.


synapseAnnotations/data/neuro.json
```diff
@@ -16,113 +16,7 @@
     "description": "",
```
karawoo: Will you add a description of what this key represents? Without the enumvalues it's not self-explanatory.

amapeters: We will be using 'group' as a key in the PhaseII PEC schema. Values will be 'institution-PI'

PR #443, opened 2019-01-22
reorganize group key

kdaily:
Moved group from `neuro` to `sageCommunity`. Also added groups from VEO-IBD and BSMN consortia.


kdaily:
Tagging @kelshmo @amapeters and @sgosline for reference.

sgosline:
I agree with the move, but why 'group' and not 'institution'? i feel like we had this discussion before, but forgot the resolution.

amapeters:
That's a good point. We use the term 'institution' in the ampad portal. The values for PsychENCODE will be 'Institution-PI'

karawoo:
At a quick glance, some of these groups have multiple institutions, like `"Columbia-SUNY"` and `"IU-JAX"`

kdaily:
The groups here are not institutions. Some are composed of individuals from multiple institutions. Some have individuals who themselves have appointments at multiple institutions (and are even listed as such on the grant). Some have groups from the same institution.

PR #450, opened 2019-02-15
Add electrophysiology assays LTP and LFP

karawoo:


allaway:
I dismissed my review because I realized at the last second that this PR is against `master`, not `develop`.

karawoo:
whoops thanks for catching that

PR #456, opened 2019-03-11
added GATK VQSR recal and tranches fileFormat

xdoan:


karawoo:
This was merged into master :(

PR #457, opened 2019-03-12
Add GATK VQSR recal and tranches fileFormats 

sgosline:


sgosline:
not sure how this ended in the wrong branch. @karawoo can you please merge?

PR #458, opened 2019-03-21
add antibody array assay

allaway:


karawoo:
We do already have assay:ELISA -- do I understand correctly that this is a more broad term, and people would use "ELISA" if appropriate, and "antibody array" otherwise?

allaway:
hmmm. I missed that. Maybe it would make more sense to add sandwich ELISA
as well, instead. The assay I am annotating is essentially a sandwich elisa
with a fancy name.

allaway:
Ok, @sgosline and @karawoo - changed this to sandwich ELISA to match the granularity of already existing terms.

synapseAnnotations/data/experimentalData.json
```diff
@@ -290,6 +290,11 @@
         "value": "LFP",
         "description": "An extracellular electrophysiology assay where a microelectrode is placed in the extracellular space of brain tissue to measure action potential and compared to an electrode either outside or inside that tissue.",
         "source": "http://purl.obolibrary.org/obo/OBI_0002189"
+      },
+      {
+        "value": "sandwich ELISA",
+        "description": "The sandwich ELISA measures the amount of antigen between two layers of antibodies. The antigens to be measured must contain at least two antigenic sites, capable of binding to antibody, since at least two antibodies act in the sandwich. So sandwich assays are restricted to the quantitation of multivalent antigens such as proteins or polysaccharides. Sandwich ELISAs for quantitation of antigens are especially valuable when the concentration of antigens is low and/or they are contained in high concentrations of contaminating protein.",
+        "source": "http://www.bioassayontology.org/bao#BAO_0002421"
```
karawoo: This link doesn't work for me?

allaway: Weird, this is either a bug with OLS or with BAO. This is the link as copied from here: https://www.ebi.ac.uk/ols/ontologies/bao/terms?iri=http%3A%2F%2Fwww.bioassayontology.org%2Fbao%23BAO_0002421

karawoo: Weird. I'd favor using the link that works even if it isn't the official URI. 

```suggestion
        "source": "https://www.ebi.ac.uk/ols/ontologies/bao/terms?iri=http%3A%2F%2Fwww.bioassayontology.org%2Fbao%23BAO_0002421"
```

allaway: I think this is caused by: https://github.com/BioAssayOntology/BAO/issues/23

So the link is technically the correct one to use for this ontology, but it doesn't actually resolve (as far as I can tell). I agree with your solution!

allaway:
OK, hopefully ready for review now.

allaway:
Thanks, @karawoo!

PR #459, opened 2019-03-22
add high content screen assay

allaway:
Like a drug screen/cell viability assay, but with a more complicated set of readouts. 
http://www.ebi.ac.uk/efo/EFO_0007550


allaway:
and besides, we just _had_ to have the word "cellomics" somewhere in this repo.

PR #461, opened 2019-03-26
Add CMC annotations

kelshmo:
The following additions are relevant to CMC.


synapseAnnotations/data/curatedData.json
```diff
@@ -38,5 +38,33 @@
         "source": "http://semanticscience.org/resource/CHEMINF_000123"
       }
     ]
+  },
+  {
+    "name": "referenceSet",
```
karawoo: Don't we already have this?

synapseAnnotations/data/experimentalData.json
```diff
@@ -565,6 +565,11 @@
         "value": "immunoassay",
         "description": "Laboratory test involving interaction of antigens with specific antibodies.",
         "source":"http://purl.obolibrary.org/obo/NCIT_C16723"
+      },
+      {
+        "value": "exonExpression",
+        "description": "Exon count values generated by DEXSeq.",
+        "source":"http://genome.cshlp.org/content/22/10/2008"
```
karawoo: If there is an existing ontology that defines exon expression I'd rather use that. If not, this is ok.

PR #471, opened 2019-05-02
added buffy coat to tissue&organ

xdoan:
Added Buffy Coat (http://purl.obolibrary.org/obo/NCIT_C84507) to tissue and organ.


amapeters:
Should be organ = blood, tissue = buffy coat?

xdoan:
I wasn't sure how far to take the "blood as an organ" with respect to the buffy coat, so blood as the organ and buffy coat as the tissue works too

amapeters:
Since a blood fraction, that to me makes most sense

Issue #482, opened 2019-05-16
Duplicate values for some keys


karawoo:
A few keys have duplicate values: 
``` r
library("tidyverse")
library("synapser")
#> 
#> TERMS OF USE NOTICE:
#>   When using Synapse, remember that the terms and conditions of use require that you:
#>   1) Attribute data contributors when discussing these data or results from these data.
#>   2) Not discriminate, identify, or recontact individuals or groups represented by the data.
#>   3) Use and contribute only data de-identified to HIPAA standards.
#>   4) Redistribute data only under these same terms of use.
library("syndccutils")
synLogin()
#> Welcome, Kara Woo!
#> NULL

dat <- get_synapse_annotations()

dupes <- dat %>%
  group_by(key) %>%
  summarize(has_dupes = any(duplicated(value))) %>%
  filter(has_dupes == TRUE)

dat %>%
  filter(key %in% dupes$key) %>%
  filter(duplicated(value)) %>%
  select(key, value) 
#>          key                             value
#> 1 fileFormat                               xml
#> 2      study                    iPSCAstrocytes
#> 3     tissue          inferior temporal cortex
#> 4     tissue medial dorsal nucleus of thalamus
#> 5     tissue     medial orbital frontal cortex
#> 6     tissue          medial prefrontal cortex
```

<sup>Created on 2019-05-16 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

The tissue issues are addressed in #480. I will fix the others in another PR. We need to investigate whether there is an issue with our integration tests, or if these just predate them, maybe. cc @kdaily

karawoo:
Running `nosetests` on the master branch (which has duplicates) does not produce any failures

karawoo:
Ok, so I think the tests didn't catch the error because the tests aren't working at all. At the time they were written, annotations were defined like so: 

```
    "libraryPrep": [
        "rRNAdepletion",
        "polyAselection",
        "lncRNAenrichment",
        "miRNAenrichment"
    ]
```

We've drastically changed the structure of our JSON files, so things like [this](https://github.com/Sage-Bionetworks/synapseAnnotations/blob/f796edd7fcf4bfd9293155123c2d65d826df016c/tests/integration/test_json.py#L22) not going to work. Also, travis tests on python 3.6 by default now, and some of whats in the tests is not compatible. So I think we need to update the tests asap.

PR #480, opened 2019-05-16
Remove duplicate tissue annotations

karawoo:
There are some duplicate values within the tissue annotation. I have removed the duplicates; all the terms are still present.

* [medial orbital frontal cortex](https://github.com/Sage-Bionetworks/synapseAnnotations/blob/f796edd7fcf4bfd9293155123c2d65d826df016c/synapseAnnotations/data/experimentalData.json#L876)
* [medial prefrontal cortex](https://github.com/Sage-Bionetworks/synapseAnnotations/blob/f796edd7fcf4bfd9293155123c2d65d826df016c/synapseAnnotations/data/experimentalData.json#L881)
* [inferior temporal cortex](https://github.com/Sage-Bionetworks/synapseAnnotations/blob/f796edd7fcf4bfd9293155123c2d65d826df016c/synapseAnnotations/data/experimentalData.json#L891)
* [medial dorsal nucleus of thalamus](https://github.com/Sage-Bionetworks/synapseAnnotations/blob/f796edd7fcf4bfd9293155123c2d65d826df016c/synapseAnnotations/data/experimentalData.json#L626)

PR #484, opened 2019-05-17
Fix test for duplicate values

karawoo:
* Fixes test so that it correctly understands the a) structure and b) location of our json data
* Updates to work with python 3.6 (which is the default on travis)

Note that it will only show an error for the first duplicate it finds. I think this was how the original test was written too, but we can update it if we want to show all duplicates at once.

This should probably fail travis checks since there are currently duplicate values until #480 and #483 are merged. See also #482.

Review on the pythonic-ness of the code is welcome, it's been a while since I wrote any python...


kdaily:
I merged #483 and restarted the build on this PR.

kdaily:
Oh. But this would have to have `develop` merged back into it.

kdaily:
will also need #486 before it works.

Issue #419, opened 2018-09-13
immunohistochemistry and brightfieldMicroscopy seem to be overlapping assays


allaway:
IHC is generally imaged via brightfield microscopy. We should consider providing guidance on when to use which term.

karawoo:
After discussion on 2018-10-09 we think the best (least-worst?) thing to do for now is guide people to use immunohistochemistry if that is relevant, and brightfield microscopy only for brightfield microscopy that is not immunohistochemistry

PR #493, opened 2019-05-30
added assay = immunofluorescence

amapeters:


karawoo:
@allaway could you weigh in on this one and whether it makes sense to have immunofluorescence in addition to immunocytochemistry and immunohistochemistry? It seems like these three terms are sometimes used interchangeably, but are really separate things, is that right? Does this also relate to #419?

allaway:
I am a little surprised we didn't already have immunofluorescence. 

From a biologist's point of view - and not an ontologist's - i think these are sufficiently different terms to merit a seperate term.
immunohistochemistry -> typically the use of immunolabeling based approaches to study one or more tissue-based proteins. These use a colored stain of some variety, and images are collected with brightfield microscopy
immunocytochemistry -> the same as above but with dissociated cells or cells from culture
immunofluorescence -> the use of fluorescent immunolabeling to study one or more proteins in cells or tissue slices, uses fluroescent microscopy. 

I think it is related to #419 in that this family of terms are not super clear or used in the same way by all people. 

Either way, I think I'd go for it, but we should probably avoid adding fluorescence microscopy as an assay to avoid confusion like the one i highlighted in #419

kdaily:
Thank you @allaway!

Followup question: what about a different term for 'Cyclic Immunoflourescence'? Asking because I have the case now that these are listed separately for MEP-LINCS.

allaway:
@kdaily I have not used CycIF myself but it sounds like it's still just immunofluorescence, but with a recycling step to reuse the sample for another epitope. if the individual epitope data (i.e. one assayed protein target per image) is provided, I'm not sure that it's worth annotating differently). If it's some sort of complicated multi-layered proprietary data type, it's might be worth labeling differently so folks can decide whether that data is worth their time...

Just my 2c - again, never worked with CycIF data.

amapeters:
Thanks for reviewing @allaway - we have all 3 as distinct assays in AMP-AD

allaway:
Ugh, I just came across an edge case here....I have data that is fluorescent microscopy, but _not_ immunofluorescence. (DAPI staining)

Should I add fluorescent microscopy as an assay?

karawoo:
I think we already have brightfield microscopy as an assay, so adding fluorescent microscopy seems reasonable to me.

allaway:
👍 thanks!

PR #494, opened 2019-05-30
Added assay = Pet and autoradiography

amapeters:


karawoo:
I'm closing this as we already have PET and autoradiography was added with #513

PR #498, opened 2019-06-07
add assays hplc-ec and whole-cell patch clamp

allaway:


synapseAnnotations/data/experimentalData.json
```diff
@@ -71,6 +71,11 @@
         "description": "An analytical technique wherein high performance liquid chromatography is coupled to tandem mass spectrometry in order to separate, identify, and quantify substances in a sample.",
         "source": "http://purl.obolibrary.org/obo/NCIT_C120691"
       },
+      {
+        "value": "HPLC-EC",
```
karawoo: I'm guessing this is not quite the same as http://purl.obolibrary.org/obo/CHMO_0001746?

It would be nice to have a source for this term that doesn't require a login to access, but if there isn't a better option then I'll approve this.

@allaway can you clarify why you've abbreviated it to `"HPLC-EC"` instead of `"HPLC-ED"`?

allaway: Agreed, login is not ideal. It was the most "reliable" source that I could find (eg. not freedictionary.com or similar...). 

At first, I thought this term was substantially different from the term you linked, but now I am second guessing myself. HPLC is not the same as LC, but HPLC-MS is frequently called LC-MS. 

I abbreviated it that way because I made a typo. :)

allaway:
I don't feel strongly either, so I'll leave it as-is. thanks for the review!

PR #503, opened 2019-06-19
Organize files in schemaorg-conversion work

karawoo:
Rearrange files as discussed in the working group. One important note ⚠️ this will make it difficult to merge `develop` back into the `schemaorg-conversion` branch to pull in new updates that have been made on `develop`. Not the end of the world, we just might have to do some cherry picking and manual cleanup. 

I have updated the README and the R script I wrote; @milen-sage would you please make sure that your code gets updated with the correct paths if needed (and double check that I deleted the right things)?


karawoo:
Actually, maybe we should merge develop first before doing this reorg. Then we'd have the updates to testing from #484, the gitignore update from #474, etc. What do you think @milen-sage?

milen-sage:
Yes - that's a good point - let's first merge develop and then merge the PR.

karawoo:
@milen-sage I merged the changes from develop into this branch. Would you mind taking a look? It's a a lot of changes so it'd be great to get another set of eyes on the results.

milen-sage:
I see  - that's great - thanks! I'll start going through later today.

PR #504, opened 2019-06-27
fileFormat additions for PEC

kelshmo:


synapseAnnotations/data/sageCommunity.json
```diff
@@ -519,6 +519,21 @@
         "value": "mtx",
         "description": "Matrix Market Exchange Format",
         "source": "https://math.nist.gov/MatrixMarket/formats.html#MMformat"
+      },
+      {
+        "value":"bfilt",
+        "description": "contains the final ATACSeq peaks after removing the encode blacklist regions.",
+        "source":"https://sites.google.com/site/anshulkundaje/projects/blacklists"
```
karawoo: Is this really a file format, or are [these files](https://github.com/Boyle-Lab/Blacklist/tree/master/lists) (are they going to be hosted on Synapse?) just BED files with specific contents?

synapseAnnotations/data/sageCommunity.json
```diff
@@ -519,6 +519,21 @@
...
+      },
+      {
+        "value":"tagAlign",
+        "description":"Tag Alignment provides genomic mapping of short sequence tags.",
+        "source":"https://genome.ucsc.edu/FAQ/FAQformat.html#format15"
+      },
+      {
+        "value":"dup",
```
karawoo: I'm guessing SAM/BAM aren't adequate for these files?

synapseAnnotations/data/sageCommunity.json
kelshmo: The file path is actually .bfilt (for example, https://www.synapse.org/#!Synapse:syn18359515)

synapseAnnotations/data/sageCommunity.json
kelshmo: Also, another strange file path .dup (https://www.synapse.org/#!Synapse:syn18359507). This output is the reprocessing pipeline for PEC - so we will continue to produce files with these file paths.

synapseAnnotations/data/sageCommunity.json
karawoo: The file format and file extension are not necessarily the same thing, so I guess I'm trying to understand what about these files warrants its own format. Is there something about the structure or format of these files that is unique and standardized?

synapseAnnotations/data/sageCommunity.json
kelshmo: tagAlign is a specific form of a bed file, dup is a SAM or BAM file with additional notation for duplicate reads...Maybe the files could be more broadly characterized, but why not be more granular and point to the specific documentation?

synapseAnnotations/data/sageCommunity.json
kelshmo: In the case of the bfilt, the file structure does not appear to deviate from a bed file so I will remove value.

synapseAnnotations/data/sageCommunity.json
karawoo: yeah that seems fine if they are different in a standardized way. my main concern was with bfilt

kelshmo:
shoot can I unmerge

PR #507, opened 2019-09-11
add type of proeomics assay

kelshmo:
If `dataType = proteomics` then it makes sense for an acceptable `assay` value to be a subtype of proteomics. In this case - TMT.

### DIFF
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -325,6 +325,11 @@
         "value": "Genotyping",
         "description": "The determination of the DNA sequence of an individual.",
         "source": "http://purl.obolibrary.org/obo/NCIT_C45447"
+      },
+      {
+      	"value": "TMT",
+      	"description":"A high-throughput multiplexed proteomics technique.", 
+      	"source":"https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0002175"
       }
     ]
   },
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -334,7 +334,7 @@
       {
       	"value": "label free MS",
       	"description":"A mass spectrometry assay that allows for measurement for endogenous targets in live cell assays, and eliminates the need for tags, dyes, or specialized reagents or engineered cells.", 
-      	"source":"https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0000708"
+      	"source":"http://purl.obolibrary.org/obo/ERO_0000708"
       }
     ]
   },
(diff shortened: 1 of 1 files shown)


amapeters:
The assay is still LC-MS/MS. Samples are TMT labeled, which is used for the spectra quantification. We do need to capture it, but something like the library prep in sequencing would, imo, be more appropriate

amapeters:
Would platform = 'tandem mass tagging' work?

amapeters:
@kelshmo - we need these annotations for AMP-AD as well. I'll contact the AMP Emory proteomics team. Let's discuss with the experts

kelshmo:
The PEC contributor confirmed this is the most appropriate naming and I followed the EMBL-EMBI [ontology](https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0002175), as noted in the schema.

kelshmo:
@karawoo  is this change in scope with your suggestion? If yes, can you add your review. Thanks!

synapseAnnotations/data/experimentalData.json
```diff
@@ -325,6 +325,16 @@
...
+      },
+      {
+      	"value": "TMT",
+      	"description":"A high-throughput multiplexed proteomics technique.", 
+      	"source":"https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0002175"
+      }, 
+      {
+      	"value": "label free MS",
+      	"description":"A mass spectrometry assay that allows for measurement for endogenous targets in live cell assays, and eliminates the need for tags, dyes, or specialized reagents or engineered cells.", 
+      	"source":"https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0000708"
```
karawoo:
```suggestion
      	"source":"http://purl.obolibrary.org/obo/ERO_0000708"
```

synapseAnnotations/data/experimentalData.json
```diff
@@ -325,6 +325,16 @@
         "value": "Genotyping",
         "description": "The determination of the DNA sequence of an individual.",
         "source": "http://purl.obolibrary.org/obo/NCIT_C45447"
+      },
+      {
+      	"value": "TMT",
+      	"description":"A high-throughput multiplexed proteomics technique.", 
+      	"source":"https://www.ebi.ac.uk/ols/ontologies/ero/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FERO_0002175"
```
karawoo:
```suggestion
      	"source":"http://purl.obolibrary.org/obo/ERO_0002175"
```

kelshmo:
I used different phrasing because it is more concise and remains unique! I can update to long form...

synapseAnnotations/data/experimentalData.json
```diff
@@ -325,6 +325,16 @@
         "value": "Genotyping",
         "description": "The determination of the DNA sequence of an individual.",
         "source": "http://purl.obolibrary.org/obo/NCIT_C45447"
+      },
+      {
+      	"value": "TMT",
+      	"description":"A high-throughput multiplexed proteomics technique.",
```
karawoo:
```suggestion
      	"description":"An isobaric labeling technique that uses tags containing four regions with the same total molecular weights and structure, so that during chromatographic or electrophoretic separation and in single MS mode, molecules labelled with different tags are indistinguishable.", 
```

Just noticed that this description differs from the source. Can we update it?

PR #521, opened 2019-09-29
Added dataSubtype = normalized

amapeters:


karawoo:
I'm a little worried that this will be confusing. "normalized" isn't mutually exclusive with the other terms here, is it? I wonder if we need a different key. Or maybe this will be a good case for multiple value annotations, when we get those.

amapeters:
@karawoo or @sgosline - please review

amapeters:
@sgosline - do you have an opinion here? We need a way to differentiate between processed, but not normalized vs processed and normalized

sgosline:
This is very use-case specific, so it's difficult to make the call. Do we still use the `dataMatrix` term? AFAIK it was  a legacy.

amapeters:
@sgosline we are not using dataMtrix. 

@karawoo we have dataSubtype - raw and processed. Processed could mean many things, but we don't have a value that differentiates between a counts matrix, and normalized expression (for example). In that case I would use processed and normalized to differentiate

synapseAnnotations/data/experimentalData.json
```diff
@@ -417,6 +417,11 @@
     "columnType": "STRING",
     "maximumSize": 250,
     "enumValues": [
+      {
+        "value": "normalized'
```
karawoo:
```suggestion
        "value": "normalized",
```

synapseAnnotations/data/experimentalData.json
```diff
@@ -417,6 +417,11 @@
...
+        "description": "A data set that is produced as the output of a normalization data transformation."
```
karawoo:
```suggestion
        "description": "A data set that is produced as the output of a normalization data transformation.",
```

amapeters:
@sgosline - can you take a look at this and let us know if you agree with using adding 'normalized' to dataSubtype

sgosline:
I agree with the term `normalized` and fully acknowledge that it's super subjective, as normalization can have it's own `json` full of terms and methods. But for now I think it's sufficient.

PR #523, opened 2019-10-04
Remove grant values

### LINKED ISSUE #519: Do we really need to control values for grant?
The grant key currently has a controlled list of values. This becomes hard to keep up to date, particularly if we get data contributions from outside the consortium in the future. Do we really need to control the values for this key, or could we change it to allow any string?

https://github.com/Sage-Bionetworks/synapseAnnotations/blob/c4c0f4f25c48271877cd596b6356891edbacf1b8/synapseAnnotations/data/neuro.json#L14-L181

karawoo:
Closes #519


karawoo:
@amapeters could you review this?

PR #528, opened 2019-10-29
Clinical data terms

cw-dvr8:
Both VEO-IBD and PsychENCODE are starting to collect clinical data on individuals, so we are proposing to add these terms as annotations.  We could use some feedback, especially on any terms that might not be appropriate to use as annotations.


cw-dvr8:
Superceded by https://github.com/Sage-Bionetworks/synapseAnnotations/pull/637

PR #538, opened 2019-11-14
Remove boolean enumerated values

### LINKED ISSUE #535: Allow alternate capitalizations of boolean values
TRUE, true, True, etc.

We are constantly having issues with people providing different capitalizations. TRUE/FALSE are the default in R and Excel. And Synapse will treat them all the same regardless of capitalization:
Screen Shot 2019-11-12 at 4 32 23 PM

So I propose all boolean annotations be updated to allow any capitalization, either explicitly as enumerated values, or implicitly by removing all the enumerated values and relying solely on the columnType. Thoughts?

karawoo:
Fixes #535

### DIFF
--- a/synapseAnnotations/data/ngs.json
+++ b/synapseAnnotations/data/ngs.json
@@ -38,19 +38,7 @@
     "name": "isStranded",
     "description": "Whether or not the library is stranded.",
     "columnType": "BOOLEAN",
-    "maximumSize": 250,
-    "enumValues": [
-      {
-        "value": true,
-        "description": "",
-        "source": ""
-      },
-      {
-        "value": false,
-        "description": "",
-        "source": ""
-      }
-    ]
+    "maximumSize": 250
   },
   {
     "name": "readPairOrientation",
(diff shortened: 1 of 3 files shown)


karawoo:
@kdaily @sgosline could one of you review this?

synapseAnnotations/data/ngs.json
```diff
@@ -38,19 +38,7 @@
...
-        "description": "",
-        "source": ""
-      },
-      {
-        "value": false,
-        "description": "",
-        "source": ""
-      }
-    ]
+    "maximumSize": 250
```
kdaily:
```suggestion
```

Think this got in there extra?

karawoo: I think it's just showing up as a changed line because I removed the comma. There is still only one instance of `maximumSize`.

karawoo: Might be easier to see in the side-by-side view

Screen Shot 2019-12-11 at 11 10 06 AM

PR #542, opened 2019-11-25
Added metadataType = quality control

amapeters:


karawoo:
@amapeters can you clarify when someone would use `metadataType: quality control` instead of `analysisType: quality control`? Is this for a particular type of metadata file you have in mind?

https://github.com/Sage-Bionetworks/synapseAnnotations/blob/c4c0f4f25c48271877cd596b6356891edbacf1b8/synapseAnnotations/data/analysis.json#L545-L547

amapeters:
I just realized that. It makes more sense to consider it an analysis

PR #547, opened 2019-12-16
Added platform = Q-Exactive Plus

amapeters:


karawoo:
@amapeters would this source work instead? http://purl.obolibrary.org/obo/MS_1002634

amapeters:
Yes, that better than the manufacturer link

amapeters:
Can you also change the link for the 'orbitrap fusion" to this http://www.ontobee.org/ontology/MS?iri=http://purl.obolibrary.org/obo/MS_1002416

karawoo:
I've updated the source. Both the manufacturer link and the new source list it as "Q Exactive Plus" (no hyphen), so I updated that as well.

PR #551, opened 2019-12-17
Add sc.rnaSeq to assays

bememea:


karawoo:
In the past (#339) I think we made the decision not to make single cell a separate assay. I don't really know enough here to weigh in -- @sgosline can you review this?

bememea:
If more context is useful, we've been getting scRNAseq data from ROSMAP and we'd like to differentiate that data from the other RNAseq data in the 'Study Data' section on the [portal](https://adknowledgeportal.synapse.org/#/Explore/Studies?Study=syn3219045).  @amapeters found that the easiest (possibly only?) way to surface the single-cell RNAseq data independently was to use an scRNAseq-specific annotation value for 'assay'.

sgosline:
Everyone is correct! In our original attempts at annotations, we followed the HCA model and used 'rnaSeq' as the assay for single cell data with the additional `nucleicAcidSource` as a value to distinguish bulk vs. single cell. This model fails with the data portal without proper aliasing logic, so Brian has started to distinguish this in the CSBC annotations. I'm ok with moving toward this direction...

karawoo:
Thanks @sgosline! Does the value `sc.rnaSeq` match what Brian has been using?

sgosline:
No, Brian uses 'Single Cell RNA-Sequencing' and also has a mapping from scRNASeq.

bememea:
sc.rnaSeq has only been used in a handful of cases, and it wouldn't be a problem at all to switch to either of the values Brian has used.

karawoo:
Hm ok, I have mixed opinions: on the one hand I'd prefer to be consistent across projects as much as possible, on the other hand "sc.rnaSeq" better matches our existing "rnaSeq" value in terms of capitalization. Anyone else have an opinion? If not I will approve this.

sgosline:
I'd argue that CSBC is so incredibly divergent from the rest of these annotations, you can go with what you see fit. I don't like the `.` in the name, and would prefer camel case or a dash (scRNASeq or sc-rnaSeq) but that is just a pet-peeve from other programming languages.

sgosline:
Again, we will adopt whatever you select.

karawoo:
Could we do `scrnaSeq`? That format would be consistent with our existing `mirnaSeq`.

PR #553, opened 2020-01-03
Added a description for the sex key

cw-dvr8:
Added a description for the "sex" key in experimentalData.


karawoo:
would you be ok with "sex assigned at birth,..."?

cw-dvr8:
I'm okay with that.  I've modified the description.

Issue #301, opened 2017-12-12
should keys have sources?


kdaily:
They have descriptions.

allaway:
I vote yes. Most will be "SAGE" but if we are going to use external descriptions (such as Organ, see #302 ) we should attribute that description to the appropriate source.

karawoo:
It seems unlikely anyone is going to have the time to go back through existing keys and add sources. Can we agree that it's ok but not required to add sources to keys?

kdaily:
I can agree!

allaway:
Nearly two year older @allaway also agrees.

karawoo:
Thanks!

PR #554, opened 2020-01-08
Add Infinium Omni2.5Exome-8 v1.3 to platform

karawoo:


cw-dvr8:
@karawoo The "platform" key is one of the ones on my list that currently doesn't have a description.  Do you feel like you can add one while you're in there, or should it be its own pull request?

karawoo:
Sure, I've added one. Does adding a source as well (per #301) interfere with anything you are doing, @cw-dvr8?

cw-dvr8:
Not at all, it's actually helpful.

karawoo:
hmm, it might be good to be specific in case there are differences between versions that affect the outputs? but i'm open to making it more general

karawoo:
@kelshmo can you confirm whether you're ok with the more specific term?

kelshmo:
I'm fine with a specific term.

PR #557, opened 2020-01-14
add two tissue terms

kelshmo:
I added two terms by request from PEC researchers. The term "insular cortex" does not have a definition but its parent does. Can we still accept this case?


cw-dvr8:
Not having a description will leave a hole in the assay templates, but otherwise wouldn't affect them. 
 Would it cause a problem to just repeat "insular cortex" as the description?

PR #559, opened 2020-02-04
Add script to update table

karawoo:
The code in annotator that updates the table of synapse annotations has been broken for a while and doesn't seem like it's going to be actively maintained. I wrote a script to update the table, thinking it might be useful to keep that within this repo where we can continue to maintain it. 

The table looks ok to me, but something seems to be broken now in the annotationUI shiny app. I've opened a Jira ticket to request access to view the logs; if there's something in this script that is causing the issue I'll fix it.

### DIFF
--- a/update_annotations_table.R
+++ b/update_annotations_table.R
@@ -45,10 +45,15 @@ unnest_module <- function(x) {
       keep_empty = TRUE
     ) %>%
     rename(
+      ## Rename columns. This works with our current JSON data, but the
+      ## numbering of columns could change if we change the underlying data
+      ## structure, so pay attention here.
       description = `description...2`,
       valueDescription = `description...6`,
       source = `source...7`
     ) %>%
+    ## Drop source column for keys because a) it's not in the synapse table
+    ## schema and b) very few keys have a source
     select(-`source...10`)
 }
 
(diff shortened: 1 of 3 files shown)


update_annotations_table.R
```diff
@@ -0,0 +1,77 @@
...
+      cols = "enumValues",
+      names_repair = "universal",
+      keep_empty = TRUE
+    ) %>%
+    rename(
+      description = `description...2`,
+      valueDescription = `description...6`,
+      source = `source...7`
+    ) %>%
+    select(-`source...10`)
```
kdaily: This is the source of the key, which currently is unused in the table. Let's comment here - maybe decide if it's useful (Kara notes only ~1 thing has a source for the key).

update_annotations_table.R
```diff
@@ -0,0 +1,77 @@
...
+## Unnest the enumerated values for each key
+unnest_module <- function(x) {
+  x %>%
+    unnest(
+      cols = "enumValues",
+      names_repair = "universal",
+      keep_empty = TRUE
+    ) %>%
+    rename(
+      description = `description...2`,
```
kdaily: This may be fragile - a change in JSON structure order may break the generated repaired names. Comment about it!

karawoo:
@kdaily will you re-review?

karawoo:
(for future reference: the app is working as of now)

PR #563, opened 2020-02-20
assay: HPLC

allaway:
Because sometimes we have HPLC without the MS!

### DIFF
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -13,6 +13,11 @@
     "maximumSize": 250,
     "enumValues": [
       {
+        "value": "HPLC",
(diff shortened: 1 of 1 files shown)


kdaily:
Thanks, @jaybee84! Our contributing guide talks a bit about that here: https://github.com/Sage-Bionetworks/synapseAnnotations/blob/master/CONTRIBUTING.md#guidelines-for-proposing-new-terms

We chose a loooong time ago NCIT as it was high quality and comprehensive for most use cases. MMO (Measurement method ontology) definitely looks useful, and if there's evidence that it is a high quality source of terms for measurements, then it could become a 'suggested' ontology to get terms from!

jaybee84:
Thanks @kdaily !

PR #564, opened 2020-02-24
Add PCRfree to libraryPrep

cw-dvr8:
Add the value "PCRfree" to the values list for the libraryPrep key.


karawoo:
Are there any ontology terms for this in the ontologies in [ontobee](http://www.ontobee.org/)? If not, I'll approve.

cw-dvr8:
@karawoo I didn't find any.  I did some Googling and found the following articles:

https://www.genengnews.com/insights/pcr-free-genome-sequencing/
https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0222562

PR #572, opened 2020-03-12
Add HI-C to libraryPrep

cw-dvr8:


karawoo:
Does "HI-C" actually describe the library prep method, or is it the assay, which we already have here? 

https://github.com/Sage-Bionetworks/synapseAnnotations/blob/5c2d86d6462b69378969976fc7dcc0e2f3b09b87/synapseAnnotations/data/experimentalData.json#L215-L217

cw-dvr8:
We have a PEC contributor who asked that it be added to libraryPrep.  Per her email:

> For template_assay_HI-C.xlsx, libraryPrep column is mandatory, while there's no relevant info I can put (currently all options are RNA-seq relevant).

> Can you tell me what you would need added to the libraryPrep column in order to make it useful to your Hi-C assay?

> HI-C would be perfect for libraryprep

I just got off of a call with @amapeters and she said she was going to investigate to see if there was a more appropriate entry.

cw-dvr8:
Found a more appropriate term.

PR #579, opened 2020-03-30
Retype chromatin amount

cw-dvr8:
I looked at EpiMap CHIPSeq data and the chromatinAmount column is a decimal.  I had it typed as an integer.  I saw by looking at other annotations modules that decimal numbers are type as STRING, so I changed the typing from INTEGER to STRING.


karawoo:
If it truly should be a number, you can make the type `DOUBLE`, e.g.: https://github.com/Sage-Bionetworks/synapseAnnotations/blob/dc25ac9d5f4375a571aec247e943b5b445e3db3f/synapseAnnotations/data/compoundScreen.json#L706-L711

Some fields like readLength are listed as strings to accommodate cases where we don't have a single value, but rather a range.

cw-dvr8:
@karawoo thanks for the clarification.  I've change the type for both antibodyAmount and chromatinAmount to DOUBLE.

PR #584, opened 2020-04-09
Additional individualIDSources for PEC

amapeters:


Aryllen:
This is out of date with the new changes to the repo (for example, this is in a folder called 'data' whereas this should be in a main folder called 'terms' and the correct subfolder). @cw-dvr8 and @amapeters, it might be simpler to close this issue and start a new one using the current version of the file.

cw-dvr8:
@Aryllen I agree.

amapeters:
@Aryllen agree as well

amapeters:
@cw-dvr8 can you make these changes

PR #588, opened 2020-04-21
Add left and right hemispheres

bememea:


karawoo:
In the past we've used "whole brain" for tissue that's not from a specific region, even if it was technically from a specific hemisphere. If we make this change, do we plan to go back and update those annotations?

bememea:
The MODEL-AD folks have asked for hemispheres specifically, since that's what they most often work with.  To be consistent within MODEL-AD I was planning to go back and reannotate entities in MODEL-AD studies that have used 'whole brain' for data generated from one hemisphere.  Any data outside of MODEL-AD that uses 'whole brain' would be left as-is.

PR #600, opened 2020-05-13
Add boolean flag for isPostMortem on specimens

Aryllen:
One use case is for iPSC where there can be a samplingAge (not currently a value in synapseAnnotations), but there needs to be an indication of whether the sample was taken post-mortem.

### DIFF
--- a/synapseAnnotations/data/experimentalData.json
+++ b/synapseAnnotations/data/experimentalData.json
@@ -1754,6 +1754,11 @@
     "columnType": "BOOLEAN"
   },
   {
+    "name": "isPostMortem",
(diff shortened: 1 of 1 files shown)


amapeters (approved):

cw-dvr8:
@amapeters Yes, ours is just called "postmortem". 

@Aryllen It looks like you're adding this to a file instead of creating a separate term for it?

Aryllen:
Thanks for catching that, @cw-dvr8! I will fix this. It also looks like this is another term that needs to be decided on (postmortem vs isPostMortem).

Aryllen:
Actually, I will close this one for now since the term is not decided and this is an old PR from before the repo changed.

PR #610, opened 2020-05-29
gitignore credentials.json

karawoo:
example_schema_usage.py downloads credentials to the directory where it's run, which would make it easy to accidentally commit it to the repo. Adding credentials.json to the .gitignore should make that less likely to happen by accident.


milen-sage:
Yes!

PR #613, opened 2020-05-29
Add documentation of the important scripts 

karawoo:
Adds documentation of the important files within the `synapseAnnotations/` subfolder. @milen-sage, there are many other scripts that I don't know the purpose of, I know you're planning to refactor some of this in the next week so I didn't go too far in trying to figure them out. One thing I did find confusing is that there's a `schemas/` subfolder, but both `sage_controlled_vocabulary_schema.csv` and `sage_schema.jsonld` are in the the `synapseAnnotations/` folder rather than in this subfolder. Should they be moved?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,17 +9,13 @@ Sage Bionetworks derived standards for annotating content in Synapse. This provi
 
 # Annotation definitions
 
-This repository contains schemas that define the things we want to annotate as well as the controlled values that may be used. You can think of these as 'columns' of a table, each with a limited set of values that can occur in each column.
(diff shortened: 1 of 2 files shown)


milen-sage:
Yes, ideally these two schema files should be under schemas (and ideally they should have the same name, up to file extension). I put them in the main directory with more explicit names, so that people can see/find them more easily initially when a limited group is testing. 

Feel free to move them, but as you mentioned there is a larger refactor comping up that will prune the unnecessary scripts/schemas and will reorganize classes into folders (i.e. python packages).

PR #624, opened 2020-07-08
PsychENCODE initial csv file

cw-dvr8:


Aryllen:
I thought the idea was to have a single csv for all of Sage, and then control the schema based on the context column. Has this changed?

cw-dvr8:
Milen hasn't implemented context yet.

Aryllen:
Oh, I see. That would be a problem. I wonder if there would be an approximate time frame for that or if some assistance would be needed.

PR #633, opened 2020-11-30
Add ad studies

Aryllen:
Changes:
- Add in 15 AD study names.
- Update MODEL-AD_APPPS1 and JAX_APPPS1 to refer to Jax.IU.Pit_APP.PS1
- Update version of study schema.

This diff makes it look like I have changed more than I have. I am wondering if it's because of the version change.


karawoo:
It looks like the diff is showing up that way because carriage returns have been added to the end of each line. This is what I see when I view the diff locally
Screen Shot 2020-11-30 at 12 54 23 PM

Aryllen:
UGH! Windows. I forgot to turn off the line ending changes for repos on this laptop. Thanks, @karawoo!

Aryllen:
That diff looks much better. 🎉

terms/neuro/study.json
```diff
@@ -305,7 +305,7 @@
         },
         {
             "const": "JAX_APPPS1",
-            "description": "The JAX_APPPS1study - A mouse model of early onset AD",
+            "description": "The JAX_APPPS1study - A mouse model of early onset AD -- now refered to as the Jax.IU.Pitt_APP.PS1 study",
```
karawoo, line 308: Is the old annotation still in use?

karawoo (commented):
Just one question about the APP/PS1 study -- if the old annotations are still in use and we're not ready to update them, then this looks good and I'll approve.

terms/neuro/study.json
Aryllen: I just checked and I don't see either of the old study names in use. My understanding was that the old values were being updated to refer to the new study name mostly for record keeping. I would not be against removing the old values altogether, but would need to double check with Mette.

terms/neuro/study.json
```diff
@@ -442,6 +442,81 @@
...
+        },
+        {
+            "const": "ROSMAP-IN",
+            "description": "The ROSMAP-iPSC derived induced neurons (ROSMAP-IN) genomic data study",
+            "source": "https://www.synapse.org/#!Synapse:syn23564891"
+        },
+        {
+            "const": "DukeAD_PTSD",
+            "description": "The Duke Study of Shared Genetic, Epigenetic, and Transcriptomic Profiles Between AD and PTSD study",
+            "source": "https://www.synapse.org/#!Synapse:syn23564891"
```
karawoo, line 509: This source goes to ROSMAP-IN

karawoo (approved):
LGTM. I think it would make sense to eventually remove the old study names, but that can happen later.

terms/neuro/study.json
Aryllen: Good catch!

terms/neuro/study.json
Aryllen: Checked with Mette and she is fine with removing the old values.

Aryllen:
@karawoo, thanks! I agree with removing old names.

PR #634, opened 2020-11-30
Fix schema registration script

karawoo:
Need to recurse through subfolders to find schemas

### DIFF
--- a/register-schemas.R
+++ b/register-schemas.R
@@ -29,7 +29,7 @@ register_schema <- function(file) {
 
 # Register terms ---------------------------------------------------------------
 
-term_files <- dir_ls(here("terms"), regexp = "*\\.json")
+term_files <- dir_ls(here("terms"), regexp = "*\\.json", recurse = TRUE)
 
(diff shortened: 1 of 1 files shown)


register-schemas.R
```diff
@@ -29,7 +29,7 @@ register_schema <- function(file) {
 
 # Register terms ---------------------------------------------------------------
 
-term_files <- dir_ls(here("terms"), regexp = "*\\.json")
+term_files <- dir_ls(here("terms"), regexp = "*\\.json", recurse = TRUE)
```
Aryllen, line 32: Are we expecting that the directories will be only a single level? If so, `recurse` could be set to an integer, which _might_ save a sliver of time depending on how the fs codebase handles recursion/checks for directories for full recursion. Limiting the level of recursion would add some inflexibility, though.

Aryllen (approved):
Overall, LGTM! One comment/thought, but approving anyway since it's not absolutely needed and potentially not desired.

PR #637, opened 2020-12-08
Demographics

cw-dvr8:
I'm proposing to add a "demographics" module to synapseAnnotations.  This is a list of demographics metadata terms that are currently in use for animal and human individuals in PsychENCODE, and also some that are in use in AMP.  These are not currently being used to annotate data files, but synapseAnnotations seemed to be a convenient central location to store them.  Please review and let me know if there are any conflicts with existing terms in other consortia.  Also let me know if I should add any other people as reviewers.


terms/demographics/ethnicity.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-demographics.ethnicity-0.0.1",
+  "type": "string",
```
Aryllen, line 4: If we are going to have an enumerated list for race, should we also have one for ethnicity?

terms/demographics/raceDetail.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-demographics.raceDetail-0.0.1",
+  "type": "string",
+  "description": "Race detail"
```
Aryllen, line 5: I don't understand what 'race detail' is. What is this term expected to be used for?

terms/demographics/reportedGender.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-demographics.reportedGender-0.0.1",
```
Aryllen, line 3: Is this needed? We have 'sex' as a term and I would expect people to use that term, regardless of whether the tissue was from a bank or not. I could see a term for 'reportedGender' if the subject was alive and was self-reporting their gender, but less so for this case.

terms/demographics/sexChromosome.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-demographics.sexChromosome-0.0.1",
+  "type": "string",
```
Aryllen, line 4: This should probably have the same enum as 'sex'.

Aryllen (commented):
Thank you for adding in the AD terms for demographics, @cw-dvr8! I do have some questions/comments and some of these may even be for terms we have in AD (I don't remember whose is whose). It would be good to nail down some of the finer details, though.

terms/demographics/Braak.json
```diff
@@ -0,0 +1,36 @@
...
+    {
+      "const": 5,
+      "source": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4894536/"
+    },
+    {
+      "const": 6,
+      "source": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4894536/"
+    }
+  ]
+}
```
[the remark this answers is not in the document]
karawoo: I agree that this would be preferable if we support min/max -- I don't know off the top of my head, but we can ask John

terms/demographics/ethnicity.json
cw-dvr8: The race categories are dictated by the NIH and have a discrete number of values.  Every ethnicity list I've found has had a large number of possible values, and are seldom consistent.  For example, some lists just list "Jewish" as an ethnicity, and some will actually break it down into separate subgroups like "Ashkenazi", "Sephardic", etc.  There are also a large number of African and Turkish/Middle Eastern ethnic identities represented in the US and elsewhere that we'd have to account for. The list would get pretty big pretty quickly and we'd probably still be missing some.

terms/demographics/sexChromosome.json
pitviper6: Is there any time when someone would need to record a sex chromosome abnormality here, in which case we'd want a string?

cw-dvr8: My understanding is that the value of this field would be an actual sex chromosome, e.g. "XX", "XY".

terms/demographics/familyHistory.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-demographics.familyHistory-0.0.1",
+  "type": "string",
+  "description": "Is there a family history of the disorder/disease",
```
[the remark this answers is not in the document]
pitviper6: I think Unknown could cause some semantic confusion, because there is a big difference between a researcher putting "unknown" because the individuals either didn't respond or weren't asked the question, and "unknown" because the person in question was adopted or for some other reason doesn't know their family history.

terms/demographics/ethnicity.json
cw-dvr8: With that said, the official governmental ethnicity list has two options:

1) Hispanic/Latino
2) Non-Hispanic/Non-Latino

If this is all we care about, we could make it an enumerated list.  This might be inadequate for future work with consortium studying diseases that are more prevalent in certain ethnic groups.

terms/demographics/ageOfDeathUnits.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-demographics.ageOfDeathUnits-0.0.1",
+  "type": "string",
+  "description": "The death age unit of measure (year or gestational week)"
```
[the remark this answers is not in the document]
amapeters: Can we set it so that if it's human we have year and gestational week, and if animal model months. I'm concerned if we have both months and year for human someone may make a mistake where 18 yrs end up as 18 months

terms/demographics/sexChromosome.json
amapeters: That is correct. For the CMC study, for example, we have a couple of XXY

terms/demographics/ageOfDeathUnits.json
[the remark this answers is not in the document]
cw-dvr8: We could specify terms such as "Human - Year", "Human - Gestational Week", "Animal Model - Month", but there is currently no way to limit the enumerated list of one term based on another term (species).  We would have to come up with two ageOfDeathUnits keys in order to do this - one human-specific for the Individual Human template and the other animal-specific for the Individual Animal template.

terms/demographics/familyHistory.json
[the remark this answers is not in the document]
cw-dvr8: I agree with Nicole's and Juliane's points.  My preference for these types of keys would be to eliminate the "unknown" option and define the key as Boolean, unless there's a specific use case to do otherwise.  "unknown" doesn't give us any more information than having a missing value in the key would.  People usually leave the column blank instead of filling in "unknown", and this causes some programming headaches for me because without a non-Boolean string value ("unknown") in the data, Python Pandas defaults true/false values internally to Boolean and I then have to take the extra step of converting them to strings in order to validate them using schemas.

terms/demographics/ageOfDeathUnits.json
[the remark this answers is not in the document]
Aryllen: I think the concern of putting 'months' depends on how people are filling these out. If they are using a neat system where there are auto-populated drop-down boxes, then I could see this being an issue since it's easy to select the wrong one. However, in that case, mistakes would be an issue for all drop-downs, not just 'ageOfDeathUnits'. If they are hand-typing or using a program to generate metadata (i.e. using a deliberate, hands-on method), this is much less likely to be a mistake that would be made.

terms/demographics/familyHistory.json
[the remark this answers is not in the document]
Aryllen: > this causes some programming headaches for me because without a non-Boolean string value ("unknown") in the data, Python Pandas defaults true/false values internally to Boolean and I then have to take the extra step of converting them to strings in order to validate them using schemas.

I have the opposite problem. I generally want booleans and having extra values is a headache. 😆

terms/demographics/sexChromosome.json
Aryllen: Ah, I see. It might be useful to clarify what is expected here by either: 
1. an enumerated list for 'XX', 'XY', 'XXY' (with future additions, if needed), OR
2. clarifying in the description what is expected (i.e.  "description": "Sex identified by SNP genotype (e.g. XX, XY)")

terms/demographics/raceDetail.json
karawoo: Agreed, I don't know what this means either

terms/demographics/Braak.json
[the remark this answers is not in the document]
Aryllen: We do not have support for minimum/maximum, yet, but there is at least one other term that was merged to production with minimum/maximum. Since we already use boolean, which is also not supported, yet, I think it would be fine to treat these as numeric with min/max and fix later, if needed (same thing we will have to do if boolean doesn't get supported).

terms/demographics/CDR.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "const": 2,
+      "description": "Moderate"
+    },
+    {
+      "const": 3,
+      "description": "Severe"
+    }
+  ]
+}
```
Aryllen, line 28: These will need a source for each. I suggest: https://doi.org/10.1017/S1041610297004870.

Aryllen (changes requested):
I 'resolved' all the boolean comments minus the main one with discussion. I still prefer sticking to boolean, especially since we already have it elsewhere in the repo. Added one more comment based on the need for sources.

terms/demographics/ageOfDeathUnits.json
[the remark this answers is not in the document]
amapeters: We will need months for mice. This isn't quite correct since for primate models year would probably be preferable ("Human - Year", "Human - Gestational Week", "Animal Model - Month",)

terms/demographics/Braak.json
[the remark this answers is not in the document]
amapeters: Since this is neuropathology, we should move this to clinical instead of demographics

terms/demographics/ethnicity.json
amapeters: I agree with @cw-dvr8 s concern about being prescriptive about ethnicity. The hispanic vs non-hispanic will probably not be sufficient for VEOIBD. Rush has a third category called Spanish ethnicity (Spanish/Hispanic/Latino origin) which is bolean. Can we adopt something similar and leave ethnicity as free text?

terms/demographics/raceDetail.json
amapeters: Is this being used in AD or PEC

terms/demographics/sexChromosome.json
amapeters: > Ah, I see. It might be useful to clarify what is expected here by either:
> 
> 1. an enumerated list for 'XX', 'XY', 'XXY' (with future additions, if needed), OR
> 2. clarifying in the description what is expected (i.e.  "description": "Sex identified by SNP genotype (e.g. XX, XY)")

I agree

terms/demographics/CDR.json
amapeters: > These will need a source for each. I suggest: https://doi.org/10.1017/S1041610297004870.

Agree. I believe this is the source Ben recomended

terms/demographics/ethnicity.json
cw-dvr8: I think that would be a good compromise.  If we ever need to do any reporting that involves the official ethnicity list, it would be pretty easy to generate Hispanic/non-Hispanic from a boolean.

terms/demographics/raceDetail.json
cw-dvr8: It's in the PEC individual human template but it doesn't appear to have any values according to the PEC clinical table.

cw-dvr8:
I've reviewed the terms that I originally had in this PR and removed the ones that I thought should actually go into the clinical module.  I'll update that PR with them.  It also means that some of the suggestions in this PR no longer apply.

terms/demographics/raceDetail.json
Aryllen: Could this definition be updated to be more descriptive of the intent? "Detailed race information"? That doesn't really add all that much clarity, though...

Aryllen (commented):
I think the only outstanding issues are clarifying descriptions on raceDetail and ageOfDeathUnits.

terms/demographics/raceDetail.json
cw-dvr8: @amapeters @kelshmo Has this key been used in another consortium where we could get more information on its use?  Since it doesn't appear that anyone has used it yet in PEC, can we just drop it?

amapeters: I think we can drop it

PR #638, opened 2020-12-09
Add diagnosisDetail

cw-dvr8:
Diagnosis subgroup (DSM or ICD code).  Schema was compiled using ajv compile -s and is valid


karawoo (approved):
This looks ok to me (I assume there are too many ICD/DSM codes for it to be worth enumerating them). Are we concerned at all that people will be mixing two different types of diagnosis codes here? If not, then this looks fine to me.

cw-dvr8:
@karawoo I'm don't know about DSM, but there are a large number of ICD codes and they are reviewed and updated on a regular basis (at least yearly, but I think more often as they are used in medical billing and in clinical trials for adverse events) so enumerating them means we'd have to commit to tracking and updating them when new terminology releases are available.

PR #639, opened 2020-12-09
Add otherMedicalDx

cw-dvr8:
Schema was compiled using ajv compile -s and is valid.

### DIFF
--- a/terms/experimentalData/otherMedicalDx.json
+++ b/terms/experimentalData/otherMedicalDx.json
@@ -1,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-experimentalData.otherMedicalDx-0.0.1",
+  "type": "string",
+  "description": "Other medical diagnoses unrelated to the primary study indication"
+}


terms/experimentalData/otherMedicalDx.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-experimentalData.otherMedicalDx-0.0.1",
+  "type": "string",
```
karawoo, line 4: If there might be multiple, perhaps we want this to be an array? I'm not sure whether JSON Schema arrays map to Synapse multi-value annotations -- if they work together harmoniously, this might be an option.

cw-dvr8: When it's been used in individual human metadata in PsychENCODE it has been entered as a string value separated by commas.  We can make it an array, but then we'd have to go back and rework existing metadata.  It is not commonly used.

karawoo: Having it as an array seems valuable for queryability, but if it's not used often then maybe it's a bit moot.

Aryllen (approved):
Assuming this is going to stay string and not change to array. String seems fine to me unless we come up with situations where string is not useful enough.

PR #641, opened 2020-12-09
Add immunoAssay keys

cw-dvr8:
Add keys to the immunoAssay module for AMP.

### DIFF
--- a/terms/immunoAssays/antibodyDilution.json
+++ b/terms/immunoAssays/antibodyDilution.json
@@ -1,0 +1,7 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "sage.annotations-immunoAssays.antibodyDilution-0.0.1",
+    "description": "Ratio of antibody used in an assay",
+    "type": "string"
+}
+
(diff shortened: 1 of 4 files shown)


terms/immunoAssays/antibodyDilution.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "sage.annotations-immunoAssays.antibodyDilution-0.0.1",
+    "description": "Ratio of antibody used in an assay",
+    "type": "string"
```
karawoo, line 5: Will people be providing this as a fraction, not a decimal?

cw-dvr8: @Aryllen would you comment on this?

Aryllen: That's a good question, @karawoo. This is expected to be a ratio like 10:100. If I remember correctly from the conversation when this was being discussed, the researchers tended to use the ratio. Should this be more clear?

karawoo: I think it's fine if it will be familiar to the researchers

Aryllen: Cool! Thank you for questioning!

Aryllen (approved):
Thank you, @cw-dvr8!

PR #642, opened 2020-12-09
Add manifold learning

cw-dvr8:
Schema was compiled with ajv compile -s and is valid.


karawoo (approved):
Wow, this analysisType key is all over the place. Some very specific analyses, and then we also have "Supervised Machine Learning". And "assessment" (???). Maybe we should rethink these values at some point, but adding this seems fine for now.

Aryllen (approved):
I have one more term added recently for this, but I can open a PR for that one. Thank you, @cw-dvr8!

PR #645, opened 2020-12-11
Add control and no cognitive impairment diagnosis

Aryllen:
Add no cognitive impairment and control to diagnosis. This one is a bit tricky since 'control' isn't technically a diagnosis. However, we have data contributors who are using this to designate control versus non-control patients.

@cw-dvr8, no cognitive impairment is in the AD annotations table so that's one you don't need to add yourself.

### DIFF
--- a/terms/experimentalData/diagnosis.json
+++ b/terms/experimentalData/diagnosis.json
@@ -135,7 +135,7 @@
         },
         {
             "const": "no cognitive impairment",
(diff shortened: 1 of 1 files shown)


Aryllen:
Forgot pathological aging. Added.

cw-dvr8 (approved):
Thanks for adding control, that's going to be very useful.

PR #647, opened 2020-12-11
Genomes1000p3

cw-dvr8:
Add 1000 Genomes phase 3 to referenceSet.

### DIFF
--- a/terms/curatedData/referenceSet.json
+++ b/terms/curatedData/referenceSet.json
@@ -6,7 +6,7 @@
         {
             "const": "1000 Genomes phase 3",
             "description": "ISGR 1000 Genomes Project Phase 3",
-            "source": "https://www.internationalgenome.org/category/phase-3/"
+            "source": "https://genome.sph.umich.edu/wiki/Minimac4#Reference_Panels_for_Download"
         },
         {
             "const": "GRCh38",
--- a/terms/curatedData/referenceSet.json
+++ b/terms/curatedData/referenceSet.json
@@ -1,8 +1,13 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "sage.annotations-curatedData.referenceSet-0.0.1",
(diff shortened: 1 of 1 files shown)


Aryllen (commented):
I am a bit confused by this. The definition of this term is:
> A set of references (e.g., canonical assembled contigs) which defines a common coordinate space for comparing reference-aligned experimental data.

1000 Genomes phase 3 seems to be referring to a whole dataset, not a common coordinate space. On the website given, the information in one of the answers is:
> 4. The reference genomes used for our alignments are different. Phase 1 alignments were aligned to the standard GRCh37 primary reference including unplaced contigs. In phase 3 we added EBV and a decoy set to the reference to reduce mismapping. This will have reduced our false positive variant calling as it will have reduced mismapping leading to false SNP calls.

The references mentioned here fit referenceSet. We already have GRCh37, but I am not sure how we would add "EBV and a decoy set".

Could I get some clarification of what this term is intended to be used for?

cw-dvr8:
@Aryllen According to the BipSeq assay description: "Genotype data were processed and normalized with the crlmm R/Bioconductor package, separately by platform. Genotype imputation was performed on high-quality observed genotypes (low-quality and rare variants were removed) using the prephasing/imputation stepwise approach implemented in IMPUTE2 and Shape-IT, using the imputation reference set from the full 1000 Human Genomes Project Phase 3 data set, separately by platform."

@kelshmo Can you comment further?

Aryllen:
@cw-dvr8, I see. I did a quick Google to see if I could find this reference set and it does exist. I see information about it (and ability to download it) in the [minimac4 documentation](https://genome.sph.umich.edu/wiki/Minimac4#Reference_Panels_for_Download). The IMPUTE2 documentation is less useful for finding the reference set information. Perhaps we just need a better source link to make this more clear?

cw-dvr8:
@Aryllen I've updated the source to the link you provided.

Aryllen (approved):
Looks good! Thanks, @cw-dvr8!

PR #650, opened 2020-12-11
CDCP consortium

cw-dvr8:
Add CDCP to consortium.  Schema was compiled with ajv compile -s and is valid.  Refer questions to @Aryllen .


Aryllen (approved):
Thanks, @cw-dvr8!

PR #656, opened 2020-12-14
Add organRIN to experimentalData

cw-dvr8:


terms/experimentalData/organRIN.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "sage.annotations-experimentalData.organRIN-0.0.1",
+    "type": "number",
+    "description": "Organ RIN, as previously measured on organ or region"
```
karawoo, line 5: I'm not clear on what this means

cw-dvr8: @karawoo If I change it to "Organ RNA Integrity Number, as previously measured on organ or region", would that clarify it?

karawoo: It's the "as previously measured on organ or region" that I don't really understand. "previously" makes it sound like maybe it has changed or become out of date? and is "region" here brain region -- if so we might want to clarify that.

cw-dvr8: I talked to @amapeters, and she said that it's asking if there is any RIN information that has been previously measured on the tissue or brain region and that it's a measure of the quality of the tissue the specimen came from that may indicate other issues.  She is going to look at how it was explained in other consortia.

cw-dvr8:
@amapeters Have you had a chance to look into how this key is being explained in other consortia?

terms/qc/organRIN.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "sage.annotations-qc.organRIN-0.0.1",
+    "type": "number",
+    "description": "Organ RIN, as previously measured on organ or region"
```
Aryllen, line 5:
```suggestion
    "description": "A general quality metric of the organ or tissue, not necessarily a direct RIN measure of the sample used in data generation"
```

Aryllen (changes requested):
Just need to update the description based on @amapeters feedback.

Aryllen (approved):
Looks good to me!

PR #659, opened 2020-12-14
Add metabolomics and pharmacokinetic assays

Aryllen:
pharmacodynamic and pharmacology are assays for dataType 'pharmacokinetic study'. The rest of the assay terms being added are for dataType 'metabolomics'. These seem like platforms/libraryPreparationMethods, but the contributors were adamant that these should be the assay names.


Aryllen:
Forgot to add 16SrRNAseq. Latest push has this one, as well. This is for dataType 'geneExpression'.

amapeters (approved):

Aryllen:
Added pharmacology as a dataType, instead, and added pharmacokinetics as an assay.

karawoo (approved):

PR #670, opened 2020-12-14
Add generic batch term

Aryllen:
We have very specific batch terms (rnaBatch, dnaBatch, libraryBatch), but we need a generic batch term.


cw-dvr8:
@Aryllen this conflicts with a "batch" term in PEC that's defined as "TMT proteomics batch".  Can we put this one on hold for a bit until we can discuss?

Aryllen:
@cw-dvr8, sure.

Our use-case for this is metabolomics. All of the other existing batch terms are too specific to a subset of assay types. There were three options that I saw:
1. We could try to come up with yet another batch term that would be specifically for metabolomics. The researchers refer to this as simply 'batch' so a new term might be confusing. It also adds to the problem of overly specific batch terms.
2. We could have a term 'batch' that is defined for metabolomics (i.e. definition: "metabolomics batch"). 'batch' is a generic word that could be used across multiple assays, but defining it specifically for a single subset of assays would, again, make this too specific a term.
3. We could have a generic batch term that would work for many assays by defining 'batch' to not be specific to a subset of assays.

We went with #3 when deciding to use this for the metabolomics templates.

amapeters (approved):

karawoo:
This should probably go into the new batch module (#695)

cw-dvr8:
I pushed the batch module to production this morning.

Aryllen:
I can update this to be in the batch module, but is the generic definition okay with you, @cw-dvr8, or do we still need to discuss?

Aryllen:
@cw-dvr8 and I agree on the generic definition. I still need to update this PR to be in the correct module.

Aryllen:
@cw-dvr8, I have updated the module. This has already been approved, but would you give it a double check before we merge?

cw-dvr8:
@Aryllen Looks good to me.

PR #671, opened 2020-12-14
New platforms

cw-dvr8:
Add biocratesP180 and BGISEQ-500 to platform.  Schema was compiled using ajv compile -s and is valid.  Refer questions to @Aryllen .


karawoo:
This looks like it has some overlap with new assays in #659, but one is spelled differently ("Biocrates p180" vs. "biocratesP180") -- I don't feel strongly about which we use, but I think we should be consistent

terms/experimentalData/platform.json
```diff
@@ -237,6 +237,16 @@
             "const": "Q Exactive",
             "description": "Thermo Scientific Q Exactive.",
             "source": "http://purl.obolibrary.org/obo/MS_1001911"
+        },
+        {
+            "const": "biocratesP180",
```
Aryllen, line 242:
```suggestion
            "const": "Biocrates p180",
```

Aryllen (changes requested):
@karawoo, that's a good point.

@cw-dvr8, I think I would lean towards "Biocrates p180" for a couple reasons (I won't list them here unless it's needed).

terms/experimentalData/platform.json
cw-dvr8: Updated.

Aryllen (approved):
Thanks, @cw-dvr8!

PR #674, opened 2020-12-15
Add tissueWeight to experimentalData

cw-dvr8:


Aryllen (approved):

terms/experimentalData/tissueWeight.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "sage.annotations-experimentalData.tissueWeight-0.0.1",
+    "type": "number",
+    "description": "Amount of tissue used to generate the specimen in mg"
```
karawoo, line 5: What does "amount of tissue used to generate the specimen" mean? Is it the weight of the specimen itself, or are the specimens created from a larger piece of tissue and that's what this annotation reflects?

cw-dvr8: @Aryllen Would you respond?

Aryllen: Weight (or rather the mass) of the specimen itself. The specimen could be a piece (or pieces) from a larger piece of tissue. 'specimenMass' would probably be more accurate of a term, but 'tissueWeight' lines up nicely with the 'tissue' type and 'weight' seems to be more understandable than 'mass'.

Would this be more easily understood as "Mass of tissue in mg" with the assumption that people will understand it's only the mass for the specimen, not the entire tissue the specimen was taken from? Other suggestions for better clarity are welcome.

karawoo: showing my ignorance here: do the specimens contain material other than the tissue itself? if not, then I'd prefer "mass of the specimen in mg". or perhaps "mass of the tissue specimen in mg" would work. but, you know this better so if you think "mass of tissue in mg" will be clear I'm fine with that.

Aryllen: > do the specimens contain material other than the tissue itself?

In general, we are assuming the material that counts as the specimen would only be from the tissue specified (and we are using tissue rather broadly in AD), but that is true that the 'tissue' itself may not necessarily relevant to what the specimen exactly is, but rather what tissue the specimen would be associated with.

I'd say this should probably be "mass of the tissue specimen in mg". This would then be both specific to the tissueWeight term and allow for someone to add a more generic term for just specimen mass, if needed.

karawoo: I'm good with "mass of the tissue specimen in mg"

Aryllen (approved):
Thanks, @cw-dvr8 for the update and @karawoo for the discussion!

terms/experimentalData/tissueWeight.json
cw-dvr8: @karawoo @Aryllen I've updated the description. Are we good to go?

PR #679, opened 2020-12-22
Remove non-ascii characters

karawoo:
Files had some non-breaking spaces and other non-ascii characters which caused issues when registering schemas on Synapse


milen-sage (approved):
I haven't had the problem of descriptions in JSONSchema failing for using greek letters, so good to know that doesn't work well with Synapse.

karawoo:
I think if you explicitly encode the body of your request as utf-8 it might work, but I haven't tried it

Aryllen:
Whoops! Thank you, @karawoo!

PR #680, opened 2020-12-22
Add message when registering schemas

karawoo:
Makes it easier to tell which schema is causing the problem if registration fails


karawoo:
Closing as superseded by #685

PR #683, opened 2020-12-23
Set up GitHub Actions to validate schemas

karawoo:
This adds a GitHub Actions workflow to check that the mini-schemas in `terms/` are 1) valid JSON Schema (using ajv-cli), and 2) contain only ASCII characters, as non-ASCII characters can cause some issues when registering schemas.

There are some other checks I'd like to add going forward, but they're a little more complicated, so I thought I'd start with this since it's useful and not too complicated.

### DIFF
--- a/.github/workflows/check-schemas.yaml
+++ b/.github/workflows/check-schemas.yaml
@@ -1,0 +1,37 @@
+on: [push]
+
+env:
+  AJV_CLI_VERSION: 4.0.1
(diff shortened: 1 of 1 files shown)


karawoo:
Note that if merged this will fail until #679 is merged.

milen-sage (approved):
Cool - can I reuse in schematic? 

Hopefully no one would want to include webdings in their key names 🕴:)

karawoo:
> Cool - can I reuse in schematic?

of course!

PR #685, opened 2020-12-24
Add workflow to check that schemas can be registered

karawoo:
This adds a GitHub Actions workflow that will test all added or changed schemas in a pull request against Synapse's schema registration endpoint with `dryRun: true`. This will ensure that schemas that are changed in a PR will be able to be successfully registered on Synapse. It does _not_ actually register them (that should probably happen in a separate action after merging the PR).

To achieve this, I've reworked the `register-schemas.R` script so it can more easily be called from the command line with a few different options. I've also set up the repo to use renv for managing dependencies, so that GitHub actions can install the correct packages. `activate.R` is auto-generated by renv and shouldn't be reviewed.

For this workflow to run, it needs Synapse credentials. I've created a service account and stored the user name and api key as secrets in the repo. PRs from forks will **not** have access to the secrets and will fail. We should discuss whether it's acceptable to have everyone make PRs from branches within the repo.

Also for discussion: the register-schemas.R script has a couple of small functions, which are currently untested. Do we want to add tests for those and, if so, do we want to set this up as a whole R package? I held off on that for now but would like to discuss more.

### DIFF
--- a/.github/workflows/register-schemas-dryrun.yaml
+++ b/.github/workflows/register-schemas-dryrun.yaml
@@ -51,5 +51,4 @@ jobs:
       - name: Attempt to register schemas with dryRun = true
         run: |
           files="${{ steps.get_file_changes.outputs.files_added }},${{ steps.get_file_changes.outputs.files_modified }}"
-          terms="$(echo $files | tr ',' '\n' | grep '^terms')"
-          Rscript register-schemas.R ${terms} --dryRun
+          echo $files | tr ',' '\n' | grep -Z '^terms' | xargs ./register-schemas.R --dryRun
--- a/.github/workflows/register-schemas-dryrun.yaml
+++ b/.github/workflows/register-schemas-dryrun.yaml
@@ -46,9 +46,10 @@ jobs:
         id: get_file_changes
         uses: trilom/file-changes-action@v1.2.4
         with:
-          output: ' '
+          output: ','
 
       - name: Attempt to register schemas with dryRun = true
         run: |
-          files="${{ steps.get_file_changes.outputs.files_added }} ${{ steps.get_file_changes.outputs.files_modified }}"
-          Rscript register-schemas.R ${files} --dryRun
+          files="${{ steps.get_file_changes.outputs.files_added }},${{ steps.get_file_changes.outputs.files_modified }}"
+          terms="$(echo $files | tr ',' '\n' | grep '^terms')"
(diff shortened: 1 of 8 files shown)


karawoo:
Note to self: this needs to be updated so that it only tries to register .json files in `terms/` -- i.e. if we had READMEs for the modules, we would not want to attempt to register them.

.github/workflows/register-schemas-dryrun.yaml
```diff
@@ -0,0 +1,53 @@
...
+          path: ${{ env.RENV_PATHS_ROOT }}
+          key: ${{ runner.os }}-renv-${{ hashFiles('**/renv.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-renv-
+
+      - name: Restore packages
+        shell: Rscript {0}
+        run: |
+          if (!requireNamespace("renv", quietly = TRUE)) install.packages("renv")
+          renv::restore()
```
tthyer, line 32: Why are the cache and restore steps necessary?

tthyer (approved):
Looks good, just had a question about cache and restore steps.

.github/workflows/register-schemas-dryrun.yaml
karawoo: The caching is just to reduce the time the job takes to run by not re-installing every R package every time. `renv::restore()` looks at the package lockfile and makes sure that the correct versions of the packages are available, and installs them if not -- so it's mainly for reproducibility of the script.

register-schemas.R
```diff
@@ -1,48 +1,123 @@
...
+  ## register_schema() can fail if e.g. schemas contain non-ASCII characters --
+  ## in this case we catch that error and provide the message.
+  token <- try(register_schema(file, dryRun = dryRun), silent = TRUE)
+  if (inherits(token, "try-error")) {
+    message(token)
+    return(FALSE)
+  }
+  ## If we get a 400 status code, synapser will throw an error -- we want to
+  ## catch that and provide the message but not raise an actual R error.
+  status <- try(get_registration_status(token$token), silent = TRUE)
```
mattfazza, line 100: I was worried about the while loop on line 67, but you're catching possible errors here.

mattfazza (approved):
Couldn't catch anything, Kara.  It looks good!  Additionally, this made me look into the different between `is` and `inherits` + `map_lgl`.  Glad to review it!

PR #686, opened 2020-12-24
Give check-schemas.yaml workflow a name

karawoo:
This will just make it a bit nicer looking in the list of workflows


Aryllen (approved):
LGTM!

PR #687, opened 2020-12-29
Use absolute paths in schema $id

karawoo:
Per [PLFM-6515](https://sagebionetworks.jira.com/browse/PLFM-6515), schemas should have absolute paths in their `$id` field. This PR updates all of the schemas' IDs, and also increments each version number (since this is a change to the schemas). `$ref`s still use relative paths.

Note that these schemas won't be able to be registered until stack-338 is deployed to prod (hence the draft PR)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ examples, but the general format for a new term should be like this:
 ```json
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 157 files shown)


karawoo:
Note to self: this needs to be updated since other changes have been merged since (so their version numbers need to be incremented again)

Aryllen:
Closing this because I've included this branch's updates in #851.

PR #689, opened 2021-01-11
directionalBSseqLibrary

cw-dvr8:
Add directionalBSseqLibrary to ngs.


terms/ngs/directionalBSseqLibrary.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "sage.annotations-ngs.directionalBSseqLibrary-0.0.1",
+  "description": "Indicates whether or not the library is directional",
```
karawoo, line 4: Does the description need to clarify that this is only for bisulfite seq, or do we think that's clear enough from the key name?

cw-dvr8: @karawoo I've clarified the description to indicate that the library is for bisulfiteSeq.

Aryllen (approved):
Looks good to me!

PR #691, opened 2021-01-12
Add sources to antibodyOrder values

Aryllen:
Add sources to antibodyOrder values. Without sources, the table script breaks.


karawoo (approved):
thank you!

PR #696, opened 2021-01-13
updated $id version to 0.0.5

avanlinden:
I forgot to update the version number when I added the HBI_scRNAseq study name


Aryllen (approved):
Looks good. Thank you for updating!

PR #700, opened 2021-01-13
Clinical terms

cw-dvr8:
Create new module clinical and add terms.


terms/clinical/CERAD.json
```diff
@@ -0,0 +1,21 @@
...
+            "const": 2
+        },
+        {
+            "const": 3
+        },
+	{
+            "const": 4
+        }
+    ]
+}
```
Aryllen, line 20: Since we need a description and source for enumerated values, this might be better as an integer type with minimum 0, maximum 4.

terms/clinical/apoeGenotype.json
```diff
@@ -0,0 +1,33 @@
...
+        {
+            "const": "34",
+            "description": "E3E4"
+        },
+        {
+            "const": "44",
+            "description": "E4E4"
+        }
+    ]
+}
```
Aryllen, line 32: These will need a source (#682). Should we just have the gene card as the source for each of them?

Aryllen (changes requested):
Two comments, one that I accidentally added without starting a review. 🤦🏻

terms/clinical/apoeGenotype.json
cw-dvr8: The gene card was originally the source for each of them, but I had thought I recalled that there had been discussion for other terms of making the individual item source into the key source if the individual item source was the same for all items.  If this is not the case, I can revert it to have the gene card on each possible value.

Aryllen: Ah, I vaguely remember that conversation now, as well. I completely agree that it makes sense as a key source. However, based on discussion about #682, we will probably require sources (unless you want to weigh in with your thoughts) for all enumerated values (this makes the dictionary apps and AD templates more useful). Due to that, the source will need to be moved back to being on the individual values.

cw-dvr8:
I've added the terms that I decided didn't belong in the demographics module.  While a few of them could have an argument made for them staying in the demographics module (e.g. genotypeInferredAncestry, sexChromosome), I ultimately decided to put them here because they are not questions that can be answered without a clinical procedure.  Am open to discussion.

Aryllen (approved):
I updated the spacing on CDR. Everything else looks good to me.

PR #702, opened 2021-01-13
New terms experimentalData

cw-dvr8:
Add new PEC/AMP terms to experimentalData


terms/experimentalData/reprogrammedCellType.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "sage.annotations-experimentalData.reprogrammedCellType-0.0.1",
+    "type": "string",
+    "description": "Cell type the specimen is modeling"
+}
```
Aryllen, line 6: @amapeters, I don't remember off the top of my head. Did we want reprogrammedCellType to have a designated set of values (for example, any value in cellType)?

Aryllen (changes requested):
One suggestion and one question.

terms/experimentalData/reprogrammedCellType.json
amapeters: Correct, any value in cellType

Aryllen: Sounds good. @cw-dvr8 and @karawoo, do you have suggestions for how we should approach this? Ideally, it would stay up to date with the values in cellType, but if we reference a version of cellType, it will not stay up to date.

karawoo: We could reference `cellType` without a version, but then `reprogrammedCellType` could not be versioned. Maybe that's ok since it sounds like we want to allow it to change automatically over time.

Aryllen: That sounds fine to me, as well.

cw-dvr8: This is similar to why I have "primaryDiagonosis" and "otherDiagnosis" in the individual metadata template both aliased to the "diagnosis" annotation in the experimentalData module.  I don't think that we should proliferate keys that are duplicates to other keys, it makes more work when one of them needs to be updated.  Are we good with dropping reprogrammedCellType as an annotation and just referencing cellType for reprogrammedCellType key in the templates?

Aryllen: @cw-dvr8, that's a good question. How about we discuss this on Friday during our meeting? I was already planning on discussing how to manage the templates for both groups. I posted some initial thoughts below.

One of the problems with dropping reprogrammedCellType is that it won't show up in the Shiny dictionary app if we do drop it, unless we tie that app to another source of information that would need to stay updated. This app is primarily what I use to find existing keys/values, and I believe Mette, Juliane, and now Abby all use it, as well. This app is also available to AD data contributors. Unless we are keeping whatever other source of information up to date in a reasonable way to maintain, then this is not as big of a problem.

Another problem is that dropping it here means that other groups within Sage may make a similar key for their iPSC work when they could have chosen to use the existing reprogrammedCellType key, had they known it existed already. Granted, they still might want their own, different key, but the option would still have been there. While I know it's improbable that we will all be on the same page for all keys/values across Sage, I still hold out hope that many will choose to reuse instead of reinvent.

Aryllen: @cw-dvr8 and I discussed this today. We were planning on adding `reprogrammedCellType` as a term since it has specific definition. However, when I went to update this term to have no version and to reference the values in `cellType`, I realized that the script for updating the table will not work correctly with this; the script isn't set up to de-refence any schemas. I would suggest pulling this term out into it's own PR.

karawoo: Argh, good point about the script, I didn't think of that.

Aryllen (approved):
Looks good to me!

PR #703, opened 2021-01-14
geneRLF

cw-dvr8:
Add geneRLF to experimentalData


Aryllen (commented):
The pdf may be better as a source instead of linked in the definition itself. However, I know some of the apps and templates we have don't expect nor show a source for keys. What do you think, @karawoo?

karawoo:
That's true, we haven't typically had sources for keys and there isn't a column for them in e.g. the annotations table on synapse. i'd lean toward leaving this as-is for now and opening a separate issue to add sources for keys and make those visible in the table.

Aryllen (approved):

PR #704, opened 2021-01-14
temperature

cw-dvr8:
Add temperature to experimentalData


Aryllen (changes requested):
This is a far better definition than what we originally had. Thanks!

The only problem I see is that this is a number, but it doesn't have a reference for the units. If we want this in Celsius, we should specify in the description that it should be Celsius (I don't see this as a problem since most of the world uses this and Fahrenheit can be converted to it). Otherwise, this would need to be a string and hope they include units.

Aryllen (approved):
Thanks, @cw-dvr8!

PR #710, opened 2021-01-25
Update AD study scRNAseq_microglia_wild_ADmice name

Aryllen:
There was a mistake in the study name, and contributor requested to change it slightly, as well.


pitviper6 (approved):
Looks good!

PR #711, opened 2021-01-26
reprogrammedCellType

cw-dvr8:
This term was pulled out of the "New terms experimentalData" PR https://github.com/Sage-Bionetworks/synapseAnnotations/pull/702 to allow the other new terms to be added because reprogrammedCellType needs more discussion.  Per @Aryllen on 25Jan:

"@cw-drv8 and I discussed this today. We were planning on adding reprogrammedCellType as a term since it has specific definition. However, when I went to update this term to have no version and to reference the values in cellType, I realized that the script for updating the table will not work correctly with this; the script isn't set up to de-refence any schemas. I would suggest pulling this term out into it's own PR."


cw-dvr8:
@Aryllen Has there been any more movement on this?

Aryllen:
Not yet. It needs to be updated to not have a version and to reference the cellType schema, but I have not done that, yet. I can update this schema later today.

The script that updates the Synapse table also needs to be updated to dereference schemas. Since the table is not being autogenerated at the moment, we could update this term and merge it, but with the caveat that if someone runs the script to generate the Synapse table, the script will fail.

I currently use part of this script when adding new terms to templates (versus relying on the table) so I've been hesitant to add reprogrammedCellType before I had time to fix the script. That's not a great reason for not adding it, though.

Aryllen:
@cw-dvr8, I updated this to reference cellType. I _think_ this is the correct way to do it based on my small tests with different ways to reference cellType.

@john-hill, could you please take a look at this before we merge it? This repository is defining a set of dictionary keys and values, with one schema per key. We want the reprogrammedCellType key to accept all the values in the cellType key. Is this an acceptable way to do this?

Aryllen:
Based on further testing, this doesn't work (in Synapse) how I had it. I updated this to have an aliased property and it appears to be working for validation.

cw-dvr8:
I was able to get Nicole's work-around with reprogrammedCellType to work with template generation.

Aryllen:
Cindy and I have both checked this out and there does not seem to be any other comments. I'm going to go ahead and merge.

Aryllen (approved):

Issue #705, opened 2021-01-15
Schemas starting with numbers


cw-dvr8:
Regarding the action item that I had from Monday's meeting for the terms "260/230" and "260/280":

I created a test schema with the name "cmolitor.test-annotations.260arrayBatch".  I compiled it using ajv compile -s and was told that the schema was valid, so it appears that the JSON schema spec accepts numbers in the schema ID.  However, when I tried to register it in Synapse, I got the following:

Invalid '$id' : 'cmolitor.test-annotations.260arrayBatch'
Encountered " <positive_digit> "2 "" at line 1, column 27.
Was expecting:
    \<letter\> ...

So this restriction is with Synapse and not JSON schema.  The options that I see are:

1. We ask John to make a change to allow schemas to start with numbers.
2. We create an alternate name for these two keys.

Note that "/" and "-" are not valid in schema IDs so we would have to modify the names of the keys anyway.

Thoughts?

Aryllen:
I'm trying to think of other cases in which we may run into this issue, but can't honestly think of any. This might simply be a weird edge case. Maybe we should just name these differently (which we have to do anyway since we can't have / or -). Hmm...ratio260over230? ratio260to230?

cw-dvr8:
The work around for the keys that were causing the issues were added to the qc modules as "ratio260over230" and "ratio260over280".  Closing this issue for now since no other keys currently start with non-alphabetic characters.

PR #712, opened 2021-01-26
QC ratios

cw-dvr8:
Originally were "260/230" and "260/280", but special characters are not allowed in schema IDs and Synapse doesn't allow schemas to start with a non-alphabetic character.  See GH Issue https://github.com/Sage-Bionetworks/synapseAnnotations/issues/705 .


Aryllen:
As you know, I am good with this. Not 'approving' since you wanted feedback from more than just us.

kelshmo (approved):
verbose work around!

PR #722, opened 2021-01-29
Add postmortem to clinical

cw-dvr8:


cw-dvr8:
I decided that this belonged in experimentalData instead of clinical because it specifically deals with tissue.  Although I suspect the possibility is pretty small, it would in theory be possible in a situation where samples were drawn at intervals (such as a clinical trial) that there could be postmortem and non-postmortem tissue samples.

Aryllen (approved):
Approving, but this is definitely another case where AD uses a similar term, isPostmortem. This is going to make tracking the dictionary terms specific to a consortium interesting...

cw-dvr8:
I didn't see "isPostmortem".  We don't need this if you've already got something, and yours sounds more descriptive anyway.  What module is it in?

Aryllen:
@cw-dvr8, I'm not sure it's more descriptive (minus the hint that it's boolean by adding 'is'). It's also not in this repo, yet, so it just comes down to opinion of which one we should have, which is subjective.

Our term: isPostMortem
definition: Was the sample was taken postmortem (true/false)?

For our term, I realized I put the capitalization wrong in the first comment. Thinking about it, this term (yours, as well) is not about the individual, but rather the sample/specimen so I would probably put this term in 'experimentalData'.

Ultimately, I think it's going to take some fancy thinking to map similar terms together (other examples like ageDeath versus ageOfDeath) so that the AD data dictionary is relevant for users. I'll give it some thought.

cw-dvr8:
I personally like yours better, so I'm going to make an executive decision and run with it.  To me the name more clearly denotes the answer we're going for, and it's not a major change that's going to throw anyone off.  You also make a good point about the module, so I'll switch that too.

Aryllen:
@cw-dvr8, alrighty. Sounds good. Still approved. Also, thank you for moving it to 'experimentalData'!

PR #723, opened 2021-01-29
Add familialRelationship

cw-dvr8:


karawoo (commented):
I like that "son" specifies that it's a human but "father" does not...

Should we consider adding other relationships besides father/son as well?

cw-dvr8:
These were the only two we required, but I can add "mother" and "daughter" as well.  I would hesitate to add any more than that until they become necessary.

karawoo:
If we do add more I think we should include gender neutral options (parent/child)

cw-dvr8:
The definitions of "child" that I found referenced a specific age group instead of a family relationship.  I added "offspring" instead.

karawoo (approved):
Thanks for the updates! Glad that github actions caught the syntax issue too.

PR #726, opened 2021-02-02
add Xevo TQ-S platform

avanlinden:


Aryllen (approved):
An extra space jumped in at the end. I pushed an update that removes it. Otherwise, looks good to me!

PR #731, opened 2021-02-10
Remove Of from key names

cw-dvr8:


Aryllen (changes requested):
I'm not sure if these have been registered, yet. If they have, then the version will need to be changed on each of these. If they haven't then this should be fine and I'll approve.

@cw-dvr8, can you please check if these have been registered?

cw-dvr8:
@Aryllen If just the name has been changed, do the version numbers need to change?  If they have been previously registered, it wouldn't have been under the new name.

Aryllen (approved):
Good point, @cw-dvr8! I was overthinking it.

PR #733, opened 2021-02-11
Delete sampleID

cw-dvr8:
This term used to be in the PEC TMT Proteomics template and was the only place it occurred.  It was never used and there's no documentation for why it was needed so it has been deleted from that template.


Aryllen (approved):
Normally, I would probably refrain from deleting terms. However, since this one was added in #702 only 16 days ago, I doubt anyone has started using it.

PR #735, opened 2021-02-11
Add sampleStatus

cw-dvr8:


Aryllen (approved):
These all need a description for the script that makes the table to function correctly. I updated this branch to have descriptions and to update the paxgene source. Please check over these to make sure they make sense, @cw-dvr8. Otherwise, LGTM!

PR #743, opened 2021-02-19
Add treatmentDose

Aryllen:
Description make sense? I have it as a string so people could include units, timeframe, etc.


avanlinden:
If it's just a string, should the description specify that units or units over timeframe should be included?

Aryllen:
How is: "Dose of treatment used, including units and timeframe"?

avanlinden:
> How is: "Dose of treatment used, including units and timeframe"?

Looks good to me!

avanlinden (approved):

PR #745, opened 2021-02-19
Add Kapa HyperPrep to libraryPrep

cw-dvr8:
I also alphabetized the entries to make them easier to scan through.


Aryllen (changes requested):
I like the alphabetization!

Only thing missing is the incrementing of the version.

Aryllen (approved):
Thanks! LGTM!

PR #746, opened 2021-02-19
Add tissues for microbiome studies

Aryllen:
Splitting hairs on whether some of these are actually "tissues," but we are using these in AD to differentiate between sample types for microbiome studies.

### DIFF
--- a/terms/experimentalData/tissue.json
+++ b/terms/experimentalData/tissue.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "sage.annotations-experimentalData.tissue-0.0.2",
+    "$id": "sage.annotations-experimentalData.tissue-0.0.3",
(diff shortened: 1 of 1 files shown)


amapeters (approved):

cw-dvr8 (commented):
I'll go along for now, but I don't like the fecal samples being in here.  Wouldn't be surprised if we don't get some pushback on this at some point if we start dealing with studies that do more standard laboratory assays.  Blood cells could be considered tissues in a stretch, but serum and urine wouldn't be.  We should probably think about coming up with a "specimenType" key, or something similar, at some point.

avanlinden:
Should "liver" be an organ, rather than a tissue?

Aryllen:
@avanlinden, liver is a term in 'organ', as well.

@cw-dvr8, agreed. This term is getting a bit polluted, most likely because it's a relatively intuitive filtering term that reduces the need for too many annotations. In the future, it might be better to create a new key.

PR #748, opened 2021-02-19
Add SentrixID and SentrixRowColumn

Aryllen:
Add SentrixID and SentrixRowColumn as new keys. These are terms from Illumina's BeadChip technology for their array-based assays. The SentrixID is the chip ID and the SentrixRowColumn is the position of the sample in the chip.

I put this in experimentalData. Should these be in a different module?


avanlinden (approved):
Both look good to me!

pitviper6 (approved):
I tried to think of why it wouldn't be in experimentalData and couldn't come up with a reason, so I think that's the right module.

PR #749, opened 2021-02-19
Add fluorphore and elisaType

Aryllen:
fluorophore seems straightforward.

elisaType relates to the ELISA assay. Technically, we have assay terms like "sandwich ELISA" that would cover this term. However, the idea for this was to be able to have an "ELISA" assay (simpler for queries) and just specify the type. Not sure if we want to keep this, though. @amapeters, thoughts on this?


avanlinden (approved):
I like having a separate elisaType key so that the assay can be just "ELISA". Looks good to me, @amapeters do you have any thoughts on this?

Aryllen:
Since this won't break anything and it gives other groups the option to use assay = ELISA with an elisaType, I'm going to merge this.

PR #754, opened 2021-02-24
Add libraryPrep

cw-dvr8:


avanlinden:
The change looks good, although I'm still unsure of the exact division between libraryPrep values and libraryPrepMethod. Since libraryPrep is the general strategy by which the library is prepared for sequencing, and a lot of the values are general methods such as polyA depletion, etc, I'm wondering if EndItDNARepairKit should go in libraryPrepMethod since it's more of specific manufacturer library kit?

@Aryllen @pitviper6 thoughts? I know we've been bumping up against this for a lot of AD terms.

Aryllen:
@avanlinden, that's a good point. These two terms are a bit confusing and, even with looking at the their definitions, it would be easy to argue that any value in one could be a value in the other. After looking at the End-IT kit information, I would probably lean more towards libraryPreparationMethod. There are a couple other values in libraryPrep that might also have been better in libraryPreparationMethod. However, that's with my idea/thoughts of how these two keys relate.

@cw-dvr8, would this value be equally useable/useful in libraryPreparationMethod or would the data require a different value there?

cw-dvr8:
@Aryllen @avanlinden I'm not really sure of the difference between these two keys.  I put this kit in here because when Kelsey, Julianne, and I had a convo over Slack about the Kapa Hyper Prep Kit, the conclusion was to add it to libraryPrep.  @kelshmo, can you weigh in here?

kelshmo:
This is my interpretation: libraryPrep is a "general strategy" and libraryPreparationMethod is a "method". A method can use a variety of different kits or, more broadly, strategies. E.g. The same method could be implemented with different kits or strategies. For example, there are several different kits that can be used to run a 10x analysis (acceptable libraryPreparationMethod value).

avanlinden:
@kelshmo Oh interesting, I think I had interpreted it in the opposite way. E.g. a "strategy" is a general approach and can use a variety of different methods or kits. But I do see your interpretation of a specific kit being a strategy for implementing a method. 

@cw-dvr8 I think it's fine to add the End-it kit here per your discussion with Kelsey and Juliane, and then maybe better delineating libraryPrep vs libraryPrepMethod would be a conversation for another AD/PEC annotation meeting?

Aryllen (changes requested):
Please update the schema version and I will approve since it seems like everyone is fine with adding this term here. There may need to be more discussion in the future to clarify the definition of these two terms, though.

Aryllen (approved):

PR #755, opened 2021-02-25
Add thalamus to tissue

cw-dvr8:
I also alphabetized it and added "Sage Bionetworks" as the source for entries that didn't otherwise have one.


Aryllen (approved):
LGTM! Thanks for the alphabetization!

PR #760, opened 2021-02-26
Shorten definition of fluorophore

Aryllen:
Max length of descriptions for keys/values is 250 in the table generated from the terms. Simplified description to get under that limit.


cw-dvr8 (approved):
Looks good!

PR #762, opened 2021-03-01
Add value to platform and alphabetize

cw-dvr8:
Added Illumina_Infinium-HumanMethylation450-BeadChip, alphabetized, and added Sage Bionetworks as a source when no other source was provided.

There are duplicate entries that I will open an issue about.


Aryllen (approved):
LGTM!

PR #764, opened 2021-03-03
Fix odd spacing character

Aryllen:
I thought I had removed all the odd spacing characters, but missed one. Spacing _should_ be fine now.


cw-dvr8 (approved):

Aryllen:
Hmm, since this is now on the main repo (not a fork), I can see that there is an error with resolving the reference. Looking into it.

Aryllen:
Okay, found the problem. The GH Action is using ajv to check that the schemas can be compiled. It's only comparing it with the other schemas that have been passed in. Since the reference to cellType does not have a version but the cellType schema does, ajv isn't finding the un-versioned cellType. Will open an issue to find a work around.

PR #766, opened 2021-03-03
Table script update: allow referencing values of a different term

Aryllen:
There are cases in which we want a term with a specific definition, but that has the same enumerated list of values as a different term. The example case is [reprogrammedCellType](https://github.com/Sage-Bionetworks/synapseAnnotations/blob/master/terms/experimentalData/reprogrammedCellType.json), which should be in sync with cellType.

This script update allows for doing this. When the table is generated, every value in cellType has two rows: one with the key cellType and the cellType definition, one with the key reprogrammedCellType and the reprogrammedCellType definition.

Assumptions/restrictions:
- Does not allow for extending terms. This will only allow for referencing the values in **one** other schema, with no additional values.
- Reference is done via `property`, where the schema term (e.g. reprogrammedCellType) is aliasing a referenced schema (e.g. cellType). See example schema linked above.
- Schema is not versioned. Script would still work, but cannot register a versioned schema with an unversioned reference.
- Referenced schema not versioned (i.e. always points to the latest version). This script would still work, but cannot register an unversioned schema that has a versioned reference.

### DIFF
--- a/update-annotations-table.R
+++ b/update-annotations-table.R
@@ -54,7 +54,7 @@ add_enumerated_values <- function(data_table, json_list) {
       mutate(value = as.character(value))
   return_df <- full_join(data_table, enum_df, by = character())
   ## If no value had a source, then need to add this column
-  if (!("source" %in% names(dat))) {
+  if (!("source" %in% names(return_df))) {
     return_df <- return_df %>%
       add_column(source = NA)
   }
--- a/update-annotations-table.R
+++ b/update-annotations-table.R
@@ -20,7 +20,11 @@ library("here")
 #' @return The module or key name
 get_info <- function(id, info = c("module", "key")) {
   info <- match.arg(info)
-  pattern <- "^[^-]+-([[:alnum:]]+)\\.([[:alnum:]]+)-[0-9\\.]+"
+  if (grepl("-[0-9\\.]+", id)) {
+    pattern <- "^[^-]+-([[:alnum:]]+)\\.([[:alnum:]]+)-[0-9\\.]+"
+  } else {
+    pattern <- "^[^-]+-([[:alnum:]]+)\\.([[:alnum:]]+)"
+  }
   switch(
     info,
     module = gsub(pattern, "\\1", id),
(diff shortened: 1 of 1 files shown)


update-annotations-table.R
```diff
@@ -20,17 +20,57 @@ library("here")
 #' @return The module or key name
 get_info <- function(id, info = c("module", "key")) {
   info <- match.arg(info)
-  pattern <- "^[^-]+-([[:alnum:]]+)\\.([[:alnum:]]+)-[0-9\\.]+"
+  if (grepl("-[0-9\\.]+", id)) {
+    pattern <- "^[^-]+-([[:alnum:]]+)\\.([[:alnum:]]+)-[0-9\\.]+"
+  } else {
+    pattern <- "^[^-]+-([[:alnum:]]+)\\.([[:alnum:]]+)"
+  }
```
karawoo, line 27: I think you could make the version part optional in the regex and avoid having to branch here, but this works too.

update-annotations-table.R
```diff
@@ -45,23 +85,23 @@ create_rows <- function(file) {
   ## If term has enumerated values in anyOf, fromJSON will create a handy data
   ## frame of those. We use that in combination with return_df to create a row
   ## for each value.
+  ## If term references the enumerated values of another term, grab those. Term
+  ## cannot extend the values in a referenced schema. Schema referenced should
+  ## exist in current set of.
```
karawoo, line 90: I think this might be missing some text at the end of the sentence

karawoo (commented):
I think this looks good but I haven't had time to try it out locally and probably won't get to that today (hence my submitting this review as a "comment"). But it looks reasonable and like it will work for our current schemas. I do think it's possible that schemas will become more complex and require a more flexible, possibly recursive approach to finding references. But that's a bridge to be crossed when we come to it, I think.

update-annotations-table.R
Aryllen: My train of thought apparently stopped right there. 😆 Thanks for always checking the comments! I'll update that later.

update-annotations-table.R
Aryllen: With the requiring of versions (see reprogrammedCellType discussion), I can remove the extra branch logic now. You're totally right that the branching would be unnecessary with an update to the regex pattern, though.

Aryllen:
Recent changes:
- Changed getting of module/term name back to the original way since all mini-schemas are versioned now.
- Finished the comment that was missing words...
- Found a bug where I was referencing the wrong variable. Not sure how I didn't run into that problem before.

@avanlinden and/or @cw-dvr8, if either of you get a chance, could you please try out this code? You don't need to run all of it. You can skip reticulate and logging into Synapse. Just need to be able to get back the correct table in this line: https://github.com/Sage-Bionetworks/synapseAnnotations/blob/d5c41d1a9dd91e8c2a03e7b74d7d57851ee5245e/update-annotations-table.R#L128. Will need to run from your local synapseAnnotations directory (the `here` package should print where it thinks "here" is when you load it).

If neither of you are able to test that this is working correctly, then I'll see if there's someone else that can try it out (or ask the Code Review group).

avanlinden:
@Aryllen When I run the line you flagged I get back the full dictionary table. I checked and there are 38 values for cellType and the same 38 for reprogrammedCellType. Seems like it works to me!

avanlinden (approved):

Aryllen:
Thank you for trying this out, @avanlinden!

PR #767, opened 2021-03-05
Add version to reprogrammedCellType and its reference

Aryllen:
This has been a 'fun' term. Originally, the idea was to not version this term so that it would stay in sync with whatever cellType had, without having to change this term, as well. However, this reduces the ability to use versioning to control what values are available to a downstream consortia schema. While it is more finicky to have to update reprogrammedCellType any time cellType is updated, the added benefit outweighs the annoyance.

This PR adds:
- a version to reprogrammedCellType $id. Note that I put this at 0.0.1 since it had no version previously.
- the most recent version to the referenced cellType $id.


cw-dvr8 (approved):
Looks as good as it's going to.

Aryllen:
And all the checks passed! Woohoo!

PR #771, opened 2021-03-11
add new term for treatmentDuration

avanlinden:
This adds a term for treatmentDuration that is a free string (meant to accompany treatmentType and treatmentDose).


cw-dvr8 (commented):
Out of curiosity, what is the reasoning behind putting it in neuro?  Most of these treatment groups were put in place before my time and I'm still trying to figure out how some terms got put into them.

avanlinden:
@cw-dvr8 My only reasoning was that the terms for treatmentType and treatmentDose are also in neuro...so not a great reason! Open to moving all of them if needed.

cw-dvr8 (approved):
No worries, that's basically how I do it too, I was just wondering if you had any other insights.

PR #775, opened 2021-03-11
Add NCX NES to tissue

cw-dvr8:


avanlinden:
If these are stem cells, should this be under cellType rather than tissue?

cw-dvr8:
Possibly, but they're listed under a brain region in a PEC study so I need them added to tissue.

Aryllen:
@cw-dvr8, would it possible to list the respective tissue for the brain region and then include cellType to drill down to these stem cells? It seems odd to have a specific cell type in tissue.

cw-dvr8:
@Aryllen No, NCX NES is what's listed as the tissue.

cw-dvr8:
However, I can add cellType and just leave tissue blank for this one.  That's probably the better solution.  Will close this and open one for cellType.

PR #777, opened 2021-03-11
Ncx nes cell type

cw-dvr8:


cw-dvr8:
@Aryllen This is failing because of the connection with reprogrammedCellType.  What do we need to do to resolve it?

Aryllen:
@cw-dvr8, update reprogrammedCellType to reference your new version of cellType. So reprogrammedCellType version should be incremented and the reference to cellType within it should be incremented. The downsides to trying to keep two terms in sync...

cw-dvr8:
@Aryllen Looks like there's a problem.  Since cellType v5 doesn't yet exist, reprogrammedCellType is failing.  Without the updated reprogrammedCellType, the PR won't pass.

Aryllen (approved):
@cw-dvr8, ugh, I forgot that attempting to register the schemas would fail because they are both being updated at the same time. 🤦🏻 It's passing the check to see that they are valid, in general (not on Synapse). Going to approve/merge!

PR #778, opened 2021-03-12
Add new values for modelSystemName and correct hAbetaK1

avanlinden:


cw-dvr8 (approved):

avanlinden:
The failing check is because I pushed from my fork, my bad.

PR #781, opened 2021-03-17
add aspen sanichip value for bedding

avanlinden:


avanlinden:
I apologize for the bracket spacing, I cannot for the life of me get it to line up (I make an edit and then it's different after I commit the changes).

pitviper6:
@Aryllen Yeah, this came out of a call where we verified that "aspen sanichip" was a specific, separate thing.

avanlinden:
@Aryllen Thank you, for single value adds I've just been making edits in the GH UI and I have never encountered a tabbing problem like that before. I thought I was going crazy, I have no idea what caused it. I usually don't use a text editor because if I submit PRs from my fork I run into the secrets issue.

PR #782, opened 2021-03-18
change waterpH key to string

avanlinden:
This changes the metadata key type to string and updates the description to specify that the value should be entered as a number on the pH scale if known, or as acidic, neutral, or basic if exact pH is not known. These changes were requested by contributors and we don't annotate with this value so it doesn't need to be controlled.

### DIFF
--- a/terms/neuro/waterpH.json
+++ b/terms/neuro/waterpH.json
@@ -1,8 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "sage.annotations-neuro.waterpH-0.0.1",
-  "description": "pH of water fed to the animal",
-  "type": "number",
-  "minimum": 0,
-  "maximum": 14
+  "$id": "sage.annotations-neuro.waterpH-0.0.2",
+  "description": "The pH of the water fed to the animal. Should be a number from 0-14 or a string specifying 'acidic', 'neutral', or 'basic'.",
+  "type": "string"
 }


cw-dvr8 (approved):

Aryllen:
I'm slightly concerned about whether this would cause a downstream problem. As things are now, if a person uploads a manifest with a column of numbers, Synapse _should_ annotate those with type integer or double. Similarly for a column of strings, it _should_ annotate with type string. If this column could be either numbers or strings, but must always be a type string to be valid, then this would make numeric annotations of waterpH be invalid.

Aryllen (changes requested):
Adding a request change to block this for a moment. @avanlinden, let's talk about this.

avanlinden:
@Aryllen Would it not just work to have the numerals considered as strings? e.g Synapse would treat the value 7.3 as "7.3"? We don't annotate with it so it's just in the metadata files. 

If this doesn't work we can come up with another key or strategy for when contributors don't have an exact pH value. @pitviper6 Can you remember which study needed this change?

Aryllen:
@avanlinden, AD doesn't annotate with waterpH, but this set of terms is available for anyone, inside or outside, of Sage. With annotation schemas available in Synapse, I think it's important to ensure that we aren't adding things that will break annotation validation for others.

pitviper6:
@avanlinden @Aryllen It's the model-ad microbiome study https://github.com/Sage-Bionetworks/AD-DCC/issues/622

avanlinden:
@Aryllen Ok good point! I will close this and take the conversation back to AD-DCC to figure out how we want to better capture this information.

Aryllen:
@avanlinden, I don't think it needed to be closed, yet, but rather, just a test done to make sure it works how it should. If it doesn't, it would be great to figure out how to get it to function correctly.

avanlinden:
@Aryllen Ok! We're also discussing just having contributors include these kind of broad pH ranges in a "notes" column.  Will leave this open for now and do some tests when I get time.

avanlinden:
I did some tests with data annotated with 'waterpH' and it does seem like Synapse is pulling the type from the manifest. When files had a mix of numeric and string values for pH, the annotation type was "text". When files had only numeric values, e.g. 7.8, 3.2, the annotation type was "floating point".

Aryllen:
@avanlinden and I spoke offline about this. Going to leave as is for now and suggest contributors to give an average value versus an ambiguous, generalized term. Can revisit if this becomes a bigger problem.

PR #783, opened 2021-03-22
add new UCI study values

avanlinden:
Study values are for MODEL-AD and don't have longer names. "The X study (X)" is pretty redundant for description but this appears to be how we did it previously for UCI studies.

### DIFF
--- a/terms/neuro/study.json
+++ b/terms/neuro/study.json
@@ -541,22 +541,22 @@
         {
             "const": "UCI_TREM2",
             "description": "The UCI_TREM2 Study (UCI_TREM2)",
-            "source": "https://www.synapse.org/#!Synapse:syn25316486"
+            "source": "https://adknowledgeportal.synapse.org/Explore/Studies/DetailsPage?Study=syn25316486"
         },
         {
             "const": "UCI_ABCA7",
             "description": "The UCI_ABCA7 Study (UCI_ABCA7)",
-            "source": "https://www.synapse.org/#!Synapse:syn25316706"
+            "source": "https://adknowledgeportal.synapse.org/Explore/Studies/DetailsPage?Study=syn25316706"
         },
         {
             "const": "UCI_ABi3",
             "description": "The UCI_ABi3 Study (UCI_ABi3)",
-            "source": "https://www.synapse.org/#!Synapse:syn25316776"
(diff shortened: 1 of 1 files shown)


amapeters:
> Study values are for MODEL-AD and don't have longer names. "The X study (X)" is pretty redundant for description but this appears to be how we did it previously for UCI studies.

I recommend writing a more comprehensive description - For example a better description for "UCI_5XFAD" study could be: 'Deep phenotyping data on the 5XFAD model from the MODEL-AD UCI team.' We can ask Alex to rewrite existing minimal descriptions

avanlinden:
@amapeters Are all of the UCI studies "Deep phenotyping data on the X model" type studies? I don't have any info on study descriptions apart from what @pitviper6 puts in the data release issue.

amapeters:
> @amapeters Are all of the UCI studies "Deep phenotyping data on the X model" type studies? I don't have any info on study descriptions apart from what @pitviper6 puts in the data release issue.

No, not always. To make it simpler we can put a short description of the model (based on the Jax description). Using 5XFAD as an example: "Data from a familial Alzheimer's disease model expressing mutant human APP and PS1, generated by the MODEL-AD UCI team"

avanlinden:
@amapeters Are these descriptions available on the Jax website? I used their mouse search tool but when I search, for example, 'TREM2', there are a ton of strains that come up and I'm not sure what to use without more information about the study.

Would it make sense to add these values as-is now so they can be used in the validator, and edit them when we have more information from the study contributors? Or should we hold off on adding until we have a better description? Sorry if I'm not understanding, I'm totally happy to add better descriptions but I just want to make sure I'm using the right info.

amapeters:
> @amapeters Are these descriptions available on the Jax website? I used their mouse search tool but when I search, for example, 'TREM2', there are a ton of strains that come up and I'm not sure what to use without more information about the study.
> 
> Would it make sense to add these values as-is now so they can be used in the validator, and edit them when we have more information from the study contributors? Or should we hold off on adding until we have a better description? Sorry if I'm not understanding, I'm totally happy to add better descriptions but I just want to make sure I'm using the right info.

We need the Jax stock number to find the exact mouse. But, yes. Creating the study description once we have it from the contributor will make this easier. A place holder is fine

amapeters (approved):

PR #787, opened 2021-03-26
Cell hashing metadata terms

avanlinden:
Per the AD/PEC discussions, I've added the following terms for a new cell hashing/CITEseq/MULTIseq assay metadata template:
- added a key for libraryType to distinguish between the hashtag oligo or barcode libraries and the cDNA transcript libraries
- added a key for sampleBarcodeID to capture which barcode corresponds to each sample 

I also proposed changing the key validBarcodeReads to specify that it applies to the **cell** barcode and not the **sample** barcode for cell hashing studies. I'm not attached to this if people think it's not necessary or there's a better way to handle this.

We may also want to discuss if want people to provide the sample barcode sequence, or just a lab-determined identifier (e.g. Barcode 1, BC-1, whatever) that is then linked in the methods description or a separate file to the nucleotide sequence of that barcode. 

Submitting this as a draft for now to solicit feedback, especially on the key-value pairs for libraryType, before I PR.


terms/qc/validCellBarcodeReads.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema",
-    "$id": "sage.annotations-qc.validBarcodeReads-0.0.1",
+    "$id": "sage.annotations-qc.validCellBarcodeReads-0.0.2",
```
cw-dvr8, line 3: If you're changing the name of the key, you don't have to update the version number because the term will have never been registered under the new name.

Aryllen:
Instead of changing validBarcodeReads, could you just add a new term called validCellBarcodeReads?

avanlinden:
@Aryllen You're right, that would probably be simpler. Do you think it would be confusing for the column in the rnaSeq and scRnaseq templates to be "validBarcodeReads" and the cell hashing template to be "validCellBarcodeReads", and they both refer to the same thing? Or do you think it would be sufficient to change the definition of the original term to clearly refer to the cell barcode, and just leave it as is?

Aryllen:
@avanlinden, looking at the original definition of validBarcodeReads, it seems like it already specifies that it's for the cell: "Fraction of reads with **cell** barcodes that match the whitelist." I'm not sure I understand the problem.

On a side note, @cw-dvr8 has been alphabetizing existing terms. It would be great if you could alphabetize libraryType before it's added.

pitviper6:
>We may also want to discuss if want people to provide the sample barcode sequence, or just a lab-determined identifier (e.g. Barcode 1, BC-1, whatever) that is then linked in the methods description or a separate file to the nucleotide sequence of that barcode.

My question on this one:

How difficult would it be for the data submitter to get the sample barcode sequence? AKA, do the labs usually create a lab-determined identifier and forget the sample barcode sequence from then on, making them do some backwards detective work?

It would be extra work for us to set up the links to the methods description or separate file (if we go that route, we should probably create a central file with the nucleotide sequences so that they can be linked to from any study that uses them). Or perhaps we could incorporate links to records in NCBI Nucleotide?

avanlinden:
@pitviper6 Good question -- it would definitely be more work to set up some kind of mapping between lab-provided barcode IDs and the actual barcode sequence, although I like the idea of a centralized document or links to NCBI Nucleotide. 

My main concern with having barcodeSequence be a key in the metadata file is just the potential for typos and copying errors in strings of 8-20ish nucleotides that appear in different places throughout the table. However, we could potentially account for this with manual validation checks when curating cell hashing data, so maybe sequence would be preferred after all.

pitviper6:
@avanlinden ooh, excellent point - there would def be typos. How onerous would manually checking all those be?

cw-dvr8:
I'm not sure that we want to get into the business of manually checking data.  It's just as easy to miss a typo as it is to make it.

avanlinden:
@cw-dvr8 I was mostly thinking just checking that the number of distinct sequences in the metadata file matches however many barcodes they say they use in the assay description -- if they use 8 barcodes but there are 9 distinct sequences in the file then flag that and have the contributor sort it out. But I agree we don't want to be actually checking data files or combing through sequence strings.

cw-dvr8:
Okay, a sanity check is always good.  However, I'd rather see checks automated whenever possible.  Write a script to count sequences in the file?

avanlinden:
Oh yeah definitely automate it. Potentially could add a function that counts unique sequences to dccvalidator that runs if you're trying to validate a cell-hashing template? We'd have to get some other kind of information from the contributor about how many barcode sequences there are supposed to be, though...

The other option is to just require that they include a small table or list with the sequences and a barcode ID for each one in the assay description or as a separate metadata file.

kelshmo:
Barcodes can be reused in a single experiment between different individuals/samples thus the user needs to see a mapping between individual/sample - batch - barcode to map the data to metadata.

I guess I am not a proponent of another lab-determined identifier that could further complicate this mapping.

kelshmo (changes requested):
My suggestion is to remove `sampleBarcodeID` and let the user provide the barcode sequence directly.

avanlinden:
@kelshmo Sounds good, I appreciate your perspective on this.

avanlinden:
I reverted the changes to validBarcodeReads and changed sampleBarcode to be a free string that should correspond to the sample barcode sequence. This is what we did for the [ROSMAP nucleus hashing metadata](https://www.synapse.org/#!Synapse:syn22415704) in the AD Portal. 

I really appreciate the feedback, everyone! Any other thoughts or issues?

kelshmo (approved):
Looks great! Did the new `libraryPrep` values get added in another PR that I missed?

cw-dvr8 (approved):

Aryllen (approved):

PR #788, opened 2021-03-29
Create chromosome key

avanlinden:
This creates a term for annotating with chromosome number. Format is an enumerated string, with values "1", "2", ... "22" + "X" and "Y". This can be updated if/when we get data from other species with different chromosome numbers and different sex chromosomes. 

I used "Sage Bionetworks" for the source for all values since chromosome number is pretty self-explanatory, but can change this if needed. 

Also open to making the values "chr1", "chr2", etc if that's more in line with what is typically used.

### DIFF
--- a/terms/experimentalData/chromosome.json
+++ b/terms/experimentalData/chromosome.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "sage.annotations-experimentalData.chromosome.json-0.0.1",
+  "description": "A structure composed of a very long molecule of DNA and associated proteins (e.g. histones) that carries hereditary information; number or designation of the particular chromosome that the data is associated with.",
+  "anyOf": [
+    {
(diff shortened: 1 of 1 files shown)


Aryllen (approved):
Looks fine to me!

cw-dvr8:
@avanlinden @Aryllen There is already a "sexChromosome" key in the PEC individual manifest.  Is this key something different?

Aryllen:
@cw-dvr8, this is for designating one specific chromosome (not sex). An example would be vcf files; it would be handy to specify which chromosome the data in the file is for.

avanlinden:
@cw-dvr8 Nicole beat me to it but that's exactly it; I think the PEC sexChromosome key is a more specific way of saying the genetically-determined sex of an individual, right? Whereas this is to let us query for data files that belong to specific chromosomes in WGS and genotype data.

cw-dvr8 (approved):
Thanks for the clarification!  Looks good to me.

PR #792, opened 2021-03-31
Add model system name values

avanlinden:


terms/neuro/modelSystemName.json
```diff
@@ -13,6 +13,11 @@
             "description": "transgenic mice overexpressing three human APP mutations and two human PS1 FAD mutations",
             "source": "https://www.jax.org/strain/008730"
         },
+        {
+            "const": "AB42",
+            "description": "",
```
cw-dvr8, line 18: We like to have something in the description, even if you're only repeating the value that's in "const".  Also, if there's not a source, we generally fill in "Sage Bionetworks" (see the "group" term under sageCommunity).

cw-dvr8 (changes requested):

terms/neuro/modelSystemName.json
avanlinden: I didn't add this value, I just moved it for alphabetizing. I think some of these are old MODEL-AD values from before we had all of our term protocols in place. I agree that this is super vague, but since there's no description or stock number I can't find this strain on the Jax website at all. Could we merge the changes adding the other values, and I can talk with @amapeters about cleaning up some of the older values without descriptions or sources?

cw-dvr8: Sounds good.  There are a lot of these and I've tried to get help cleaning them up.

cw-dvr8 (approved):

terms/neuro/modelSystemName.json
avanlinden: True. Does PEC use these at all? If not, I can start an issue in the AD-DCC metadata project to tackle going through our annotations, seeing what's being used, and cleaning things up a bit.

cw-dvr8: We didn't used to, but now that we're sharing an individual_animal template with AD we are.

PR #794, opened 2021-04-02
New uci study values

avanlinden:
- added new study names and modelSystemName values for new UCI mouse studies
- alphabetized study name list


terms/neuro/modelSystemName.json
```diff
@@ -199,7 +199,7 @@
         	"source": "https://www.jax.org/strain/31950"
         },
         {
-        	"const": "PicalmH458R_Aduci",
+        	"const": "Picalm*H458R",
```
cw-dvr8, line 202: I have some concerns with the asterisk because asterisks have special meanings at times, e.g. regular expressions.  Can it be left out?

terms/neuro/modelSystemName.json
```diff
@@ -228,6 +228,11 @@
             "description": "triple mutant strain carrying a humanized ApoE knock-in mutation, an A528T point mutation of the Sorl1 gene. and an R47H point mutantion of Trem2",
             "source": "https://www.jax.org/strain/031940"
         },
+        {
+            "const": "Spi1*rs1377416",
```
cw-dvr8, line 232: I have some concerns with the asterisk because asterisks have special meanings at times, e.g. regular expressions.  Can it be left out?

avanlinden: Thanks for catching this, didn't even think about that. I removed the asterisks from both and I'll have Juliane let the contributor know we can't use them in model names.

cw-dvr8 (approved):

PR #804, opened 2021-04-15
add controlType values

avanlinden:


terms/experimentalData/controlType.json
```diff
@@ -22,6 +32,11 @@
             "const": "study pool",
             "description": "Reference pool obtained from study samples.",
             "source": "Sage Bionetworks"
+        },
+        {
+            "const": "water plus Baker ISTD",
```
Aryllen, line 37: It seems there should be a better way to name this (just Baker ISTD?), but ultimately approving because there's nothing legitimately wrong with it.

Aryllen (approved):
Looks fine to me.

terms/experimentalData/controlType.json
avanlinden: I am open to a better name! The contributor specified that this is a "blank", i.e. water plus the standards, whereas sometimes they use serum plus the standards, so I wanted to keep "water" in case we later get someone using the serum. But I agree it's clunky.

Aryllen: I'm completely blanking on clever names. 'water-based Baker ISTD'? That's not any better than 'water plus Baker ISTD'. 😆 That makes sense to include 'water', though.

PR #810, opened 2021-05-05
add SuperAgerEpiMap study

avanlinden:


Aryllen (approved):
I see that you used the Portal link to the study. That's probably a better way to link to the studies. Might want to eventually change all the AD-study sources to the portal.

avanlinden:
@Aryllen That's the eventual plan!

PR #820, opened 2021-05-27
Add AD terms

avanlinden:


terms/experimentalData/assay.json
```diff
@@ -458,6 +458,11 @@
             "description": "Ultra performance liquid chromatography - tandem mass spectrometer.",
             "source": "https://www.wur.nl/en/show/Ultra-performance-liquid-chromatography-tandem-mass-spectrometer-UPLCMSMS.htm"
         },
+        {
+            "const": "UPLC-ESI-QTOF-MS",
```
Aryllen, line 462: Wow, that looks fun. I'm not sure if people will recognize that abbreviation. I noticed that the paper referenced a protocol with a shorter abbreviation. I feel like this is getting more into the weeds with the differences between assays; that's extremely specific. That said, I'll approve it since we have other overly specific assays in here, as well.

avanlinden: Yeah I don't love it but the contributor was clear that distilling it to something shorter would not be accurate. I think part of the problem is that it's a big mash of different machines -- the UPLC does the chromatography, the ESI sprays out the ions, and the QTOF-MS does the mass spec categorization. Hopefully this will be more common as we get more metabolomics data.

PR #822, opened 2021-06-03
Add AD study value and fix tissue type

avanlinden:


Aryllen (approved):
Thank you for checking what the existing annotations are using for capitalization!

PR #823, opened 2021-06-03
Add AD assay metadata template keys and values

avanlinden:
This PR adds terms for 6 new AD-related keys for MRI, PET, and autorad templates.

### DIFF
--- a/terms/manifestKeys/filename.json
+++ b/terms/manifestKeys/filename.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "sage.annotations-manifest.filename-0.0.1",
+  "$id": "sage.annotations-manifestKeys.filename-0.0.1",
   "description": "The name of a file.",
   "type": "string",
   "minLength": 1,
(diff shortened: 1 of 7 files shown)


Aryllen (changes requested):
I noticed that all the strings here are limited to 50 characters. This is handy because it could decrease problems with file views, if any of these end up annotations. I don't _think_ 50 characters is too limiting so it should be fine.

I think `filename` would be better in a different module, though. It's not exactly a manifestKey, but it is a file-related key and manifests are file-related. What do you think about putting this in the manifestKey folder/module?

avanlinden:
I just used the term templates for free strings -- I can change the max string length if needed. I _think_ 50 should be enough for all of them, though? Unless there's an upsettingly long contrast agent or something. 

I moved filename to the manifest module.

terms/manifestKeys/filename.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "sage.annotations-manifest.filename-0.0.1",
```
Aryllen, line 3:
```suggestion
  "$id": "sage.annotations-manifestKeys.filename-0.0.1",
```
Just one minor fix.

avanlinden: Ayyy good catch.

Aryllen (approved):
Looks good to me! The string lengths can be changed in the future, if needed.

PR #828, opened 2021-06-14
Added three values

pitviper6:
LaminB, H3K9me2, SON


terms/experimentalData/assayTarget.json
```diff
@@ -394,10 +404,15 @@
             "source": ""
         },
         {
-            "const": "SOX8",
+            "const": "SMC3",
```
Aryllen, line 407: It looks like "SOX8" was replaced with "SMC3". Did this need to be in addition, not in replacement?

Aryllen (changes requested):
One question below.

Aryllen (approved):
LGTM

PR #841, opened 2021-07-07
 invitro biospecimen terms

avanlinden:
This update adds several new terms that will be used for the AD in vitro biospecimen template, which is meant for studies that use cell lines, iPSCs, and organoids.

Enumerated keys: modelSystemType, sourceCell, reprogrammingMethod, differentiationMethod

Open string keys: stemCellBaseMedium, plateCoating

Note: sourceCell has many of the same values (although not all) as the existing term cellType. I opted not to make it an alias because many of the values in cellType are not needed for sourceCell, but I am open to changing this. 

I anticipate values being added to these terms as we get more iPSC and organoid data in the AD portal.


terms/experimentalData/differentiationMethod.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "sage.annotations-experimentalData.differentiationMethod-0.0.1",
+  "description": "The general strategy by which a reprogrammed or pluripotent cell is differentiated into a target cell.",
+  "anyOf": [
+        {
+            "const": "directed differentiation",
+            "description": "Cell differentiation via the addition of growth factors and small molecules that modulate developmental signaling pathways.",
+            "source": "https://doi.org/10.1016/j.ymeth.2015.11.019"
```
Aryllen, line 9: Should this source be the same as the Ngn2 one?

terms/experimentalData/reprogrammingMethod.json
```diff
@@ -0,0 +1,23 @@
...
+            "const": "episomal vectors",
+            "description": "Non-integrative episomal vector methods used to reprogram somatic cells into induced pluripotent stem cells.",
+            "source": "https://doi.org/10.1177/0963689719886534"
+        },
+        {
+            "const": "Sendai virus",
+            "description": "A pluripotency induction technique that employs Sendai virus vectors to force expression of a combination of genes sufficient to induce pluripotency without integration or alteration of the cellular genome.",
+            "source": "http://purl.obolibrary.org/obo/ERO_0002163"
+        }
+
```
Aryllen, line 21:
```suggestion
```
Just an extra space here. Can fix by clicking on the "commit suggestion" button.

terms/experimentalData/stemCellBaseMedium.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "sage.annotations-experimentalData.stemCellBaseMedium-0.0.1",
+  "description": "Culture medium that stem cells are grown in during differentiation.",
+  "type": "string",
+  "minLength": 1,
+  "maxLength": 50
+}
+
```
Aryllen, line 9:
```suggestion
```
It looks like there might be two newlines at the end of this file. This suggestion should remove one of the two (only one is needed for the file to open and be read correctly).

Aryllen (changes requested):
Looks pretty good to me! Just one question and a couple suggestions.

terms/experimentalData/differentiationMethod.json
avanlinden: Yeah, it's a review paper that explains both terms.

Aryllen (approved):
Looks good to me!

PR #842, opened 2021-07-09
add AD-BXD study

avanlinden:
This must have been added at one point because it's an older study but it disappeared. Re-adding.


Aryllen (approved):
That's odd that it disappeared. It's probably somewhere in the commit history of this file, but just adding it back is fine. Hopefully there are not many like this.

PR #843, opened 2021-07-09
Add new DCC values

avanlinden:


terms/experimentalData/assay.json
```diff
@@ -438,6 +438,11 @@
             "description": "SNP measurements captured by array technology",
             "source": "http://purl.obolibrary.org/obo/OBI_0001204"
         },
+        {
+            "const": "snrnaSeq",
```
Aryllen, line 442: scrnaSeq and snrnaSeq look very similar. 😆

Aryllen (approved):
LGTM!

terms/experimentalData/assay.json
avanlinden: Sure do! Fingers crossed this reduces confusion instead of increasing it... 😬

PR #844, opened 2021-07-19
New terms for STARRSeq assay

kelshmo:
The PR: 

- Adds a new key commercialSource for nucleic or amino acids purchased from a private company. The goal here is to surface an accession number. 
- Modifies the definition of `libraryType` to be slightly more permissive. For this use case, this key is to also be applicable to plasmid insert ampliciation which is pooled but not _necessarily_ multiplexed.  
Screen Shot 2021-07-19 at 10 59 07 AM
- Adds value `plasmid DNA` to `libraryType`
- Adds value `cell-free DNA` to `nucleicAcidSource`

### DIFF
--- a/terms/ngs/commercialSource.json
+++ b/terms/ngs/commercialSource.json
@@ -1,0 +1,13 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "sage.annotations-ngs.commercialSource-0.0.1",
+    "description": "Nucleic or amino acids characterized that originate from a commercial source.",
+    "anyOf": [
+        {
+            "const": "Promega G1471",
+            "description": "Commercially available human genomic DNA distributed by Promega.",
+            "source": "https://www.promega.com/products/biochemicals-and-labware/nucleic-acids/human-genomic-dna/?catNum=G1521"
+        }
+    ]
+}
+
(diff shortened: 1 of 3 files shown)


pitviper6 (commented):
Do we also want to capture the accession numbers for the amino acids?

terms/ngs/commercialSource.json
```diff
@@ -0,0 +1,13 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "sage.annotations-ngs.commercialSource-0.0.1",
+    "description": "Nucleic or amino acids characterized that originate from a commercial source.",
```
avanlinden, line 4: Would it make sense to expand this term to use for commercial cell lines or stem cells? We've previously used individualIdSource to capture the vendor name for commercial cells, but this is definitely more specific.

kelshmo: Yes, we can expand for commercial cell lines OR capture commercial products in individualIdSource. I thought individiualIdSource referred primarily to brain banks but I now realize if the dna/tissue/rna/cell line was of commercial source, there would not be redundancy with that key.

avanlinden: I would say we can leave as is for now and expand the definition to use for cell products later if we need it. Thanks for confirming!

avanlinden (approved):

Aryllen (approved):
This looks good to me. The check is failing because this is branch is from a forked repo; the GH action doesn't have access to the login secrets.

PR #849, opened 2021-07-29
alphabetize terms

kelshmo:
Since the list of assayTarget values is so long, proposing to alphabetize the values as it gets listed in this order in the metadata templates. 

Quick review, thanks @avanlinden !


avanlinden (approved):
Looks good!

PR #851, opened 2021-08-11
Use absolute paths and add missing string types

Aryllen:
This PR brings in the work from #687, updates it to include absolute paths on newer terms, and adds missing string types for enumerated values.

Includes:
- absolute id paths
- incremented versions
- `"type": "string"` for terms with a list of allowed values
- alphabetical ordering for softwareLanguage
- simple descriptions added for a few terms


Aryllen:
I've double-checked the files. The basic schema checks passed and the registration check failed for reprogrammedCellType not finding version 0.0.8 of cellType, as expected. Merging since this is blocking annotation schema testing.

PR #853, opened 2021-08-12
Do not use absolute path for cellType reference in reprogrammedCellType

Aryllen:
Absolute paths were added in #851. The absolute paths for ids is fine, but these should not be used for referenced terms.

- Change back to relative path for cellType in reprogrammedCellType


Aryllen:
Tested in Synapse. All checks should pass since cellType version did not change. Will merge once the tests pass.

PR #857, opened 2021-08-13
Use enumerated cellType list instead of reference for reprogrammedCellType

Aryllen:
The reprogrammedCellType schema was using a reference to cellType to keep both up to date with each other. This ended up being a bit of a pain due to failing CI checks when cellType is updated. Now that we have a web interface for editing annotations based on a schema, this term also looks odd in that interface.

Instead of sticking with the reference, this PR duplicates the cellType values in reprogrammedCellType. If one gets out of date with the other and is in need of updating, it is not difficult to create a PR to make the change.


pitviper6 (approved):
This is a great improvement!

PR #860, opened 2021-09-07
Add GH Action to register newly added or modified files on merge and update Synapse table

Aryllen:
A GH Action exists to test that newly added/modified files can be registered in Synapse, but the terms are not actually registered. This is a second action that would register the terms in Synapse once they are merged into the main branch. The code here is an updated version of the dryrun action.

GH Actions are tough to test before deploying, especially when they trigger on the main branch. If someone could skim this and gives a thumbs up if it looks okay, that would be great!

Edit: I also decided to throw in the table update since that should be done on merge to main, as well.

### DIFF
--- a/.github/workflows/register-schemas.yaml
+++ b/.github/workflows/register-schemas.yaml
@@ -9,9 +9,7 @@ env:
 
 jobs:
   register-schemas-dryrun:
-    runs-on: macos-10.15
-    strategy:
-      fail-fast: false
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
 
@@ -38,8 +36,7 @@ jobs:
           OUTPUT_FILE=~/.synapseConfig
           cat > "$OUTPUT_FILE" << EOM
           [authentication]
-          username = "${{ secrets.SYNAPSE_USER }}"
-          apikey = "${{ secrets.SYNAPSE_APIKEY}}"
+          authtoken = "${{ secrets.SYNAPSE_PAT }}"
           EOM
           chmod +x "$OUTPUT_FILE"
 
--- a/.github/workflows/register-schemas.yaml
+++ b/.github/workflows/register-schemas.yaml
@@ -2,7 +2,7 @@ on:
   push:
     branches: master
 
-name: Register schemas
+name: Register schemas & update table
 
 env:
   RENV_PATHS_ROOT: ~/.local/share/renv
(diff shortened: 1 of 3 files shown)


.github/workflows/register-schemas.yaml
```diff
@@ -0,0 +1,59 @@
...
+    branches: master
+
+name: Register schemas & update table
+
+env:
+  RENV_PATHS_ROOT: ~/.local/share/renv
+
+jobs:
+  register-schemas-dryrun:
+    runs-on: macos-10.15
```
thomasyu888, line 12: Any particular reason you are using macos?

.github/workflows/register-schemas.yaml
```diff
@@ -0,0 +1,59 @@
...
+        run: |
+          if (!requireNamespace("renv", quietly = TRUE)) install.packages("renv")
+          renv::restore()
+
+      - name: Set up Synapse config file for testing
+        run: |
+          OUTPUT_FILE=~/.synapseConfig
+          cat > "$OUTPUT_FILE" << EOM
+          [authentication]
+          username = "${{ secrets.SYNAPSE_USER }}"
```
thomasyu888, line 41: Use PAT

thomasyu888:
Overall, LGTM.  I know our security expert is not a huge fan of interacting with Synapse via Github CI/CD, so just on that alone, I would decline this PR.

That being said, if this doesn't contain any PHI data, then perhaps the solution here is to create a service user with extremely scoped access.

.github/workflows/register-schemas.yaml
Aryllen: Nope. It's just what Kara set the dryrun action up as. I just copy/pasted and updated that action.

zaro0508: i believe GH, just like all other CI services, has much less macs than linux.  i'm guessing that linux is much more available and much faster

.github/workflows/register-schemas.yaml
```diff
@@ -0,0 +1,59 @@
...
+name: Register schemas & update table
+
+env:
+  RENV_PATHS_ROOT: ~/.local/share/renv
+
+jobs:
+  register-schemas-dryrun:
+    runs-on: macos-10.15
+    strategy:
+      fail-fast: false
```
zaro0508, line 14: why set this to `false`? don't you want to stop on first failure?

zaro0508 (changes requested):

.github/workflows/register-schemas.yaml
Aryllen: Good catch. It will stop on first failure because it's only a single job. However, I will remove that if I stick with the GH Action since it's not needed in situations with only one job. If there were a matrix of jobs, I would not want it to fail fast because it automatically stops all other jobs in the matrix.

From the docs:
> When set to `true`, GitHub cancels all in-progress jobs if any matrix job fails. Default: `true`

Aryllen:
@thomasyu888 and @zaro0508, I've updated this action and the older action to use a PAT, Ubuntu, and fail-fast. Thank you both!

Aryllen:
Eh, nix the linux. I'd have to install more stuff and I don't have the time to play wack-a-mole for these actions. It's fine if these actions takes a moment longer while waiting for macos. The idea is that the actions run fairly soon-ish after a change. So far, there's not been a problem with them running quickly.

zaro0508 (approved):

PR #884, opened 2021-11-22
Adding several terms for MODEL-AD individual animal template

RYaxley:


terms/neuro/bodyWeight.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/sage.annotations-neuro.bodyWeight-0.0.1",
+    "type": "number",
+    "description": "Weight of animal in grams"
```
avanlinden, line 5: Is there a way to make this key apply to humans too? I guess if we specify the units in it, it would be weird to have a human weight in grams.

terms/neuro/climbID.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/sage.annotations-neuro.climbID-0.0.1",
+  "description": "Identifying string linked to an animal in Climb database",
```
avanlinden, line 4: Can we specify where the Climb database is? (i.e. hosted by Jackson Laboratories)

terms/neuro/diet.json
```diff
@@ -0,0 +1,33 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/sage.annotations-neuro.diet-0.0.1",
```
avanlinden, line 3: Would we potentially want to call this something like "animalDiet," in case we need "diet" as a key for human data at some point? Possibly the gut microbiome data?

avanlinden (commented):
I had a couple of questions about making terms generalizable, but the formatting of the JSON schemas looks correct and everything else looks good. @pitviper6 @amapeters Any thoughts on whether we would want 'bodyWeight' and 'diet' terms to eventually be used for human metadata as well, e.g. for the gut microbiome stuff?

terms/neuro/climbID.json
RYaxley: I think each lab manages its own instance of Climb. We can include a URL to the developer if helpful. https://www.rockstepsolutions.com

terms/neuro/diet.json
RYaxley: animalDiet makes sense

terms/neuro/bodyWeight.json
avanlinden: Nevermind, let's stick with this for now because the alternatives are worse.

terms/neuro/climbID.json
avanlinden: Ah, never knew that. Yeah, let's include the link to the developer.

terms/neuro/diet.json
RYaxley: Since this is term being added for MODEL-AD, would it make sense to have a more specific name like:
- rodentDiet
- mouseDiet

avanlinden: Since this will be specific to model-ad, I'll defer to you on which name would be most appropriate!

avanlinden: Since it's model-ad specific I defer to your judgement on which specific term would be best!

pitviper6 (approved):

PR #886, opened 2021-12-16
add sac-seq to libraryPreparationMethod

danlu1:


danlu1:
@pitviper6 The reason this value is not in dictionary is that it should be approved before merging to master branch. Then the following workflow of updating dictionary will be triggered. Please help to check and approve the change.

PR #890, opened 2022-01-06
Update individual animal modelad

RYaxley:


terms/neuro/individualCommonGenotype.json
```diff
@@ -3,21 +3,16 @@
   "$id": "sage.annotations-neuro.individualCommonGenotype-0.0.1",
```
avanlinden, line 3: increment version again, since this has already been registered once

avanlinden (commented):
increment version in individualCommonGenotype so it can register; otherwise looks good

avanlinden (approved):

PR #898, opened 2022-02-14
Add terms

ryaxley:


avanlinden:
Per our discussion w/ Mette, I'll close this PR and we'll provide guidance to contributors to use multiple values for race in place of "multiracial".

PR #907, opened 2022-03-24
add ELITE consortium value

avanlinden:
Add ELITE

Second attempt PR because of problems with dpulls


ryaxley (approved):
Still looks good

PR #909, opened 2022-04-05
add libraryPrep for PEC NeuMet study

danlu1:


avanlinden:
@danlu1 I think @pitviper6 mentioned this to me -- I can't tell if `methylSeq` should be an assay or a libraryPrep value. Do you know what kind of data or assay the contributor is using this with?

danlu1:
I can't tell since there is no wiki and description on GitHub. @pitviper6 Can you explain a bit?

pitviper6:
@avanlinden @danlu The data and assay are:
oxBS-Seq: 50 donors, 100 samples

    FANS of 50 samples, DNA isolation, sequencing library construction, sequencing, pre-processing, QC

When I asked for clarification on libraryPrep, the DC said this:
>For this set, I've corrected the manifest. And also add libraryPrep in the assay as "NuGEN Ultralow Methyl-seq with True methyl oxBS". I can't find any corresponding value for that. Please let me know if I need to change that.

I went to the doucmentation for the NuGEN Ultralow, and it looks like it does use methylSeq as the prep but I could be wrong.

avanlinden (approved):
Alright, looks good to me then!

pitviper6 (approved):

PR #912, opened 2022-04-28
add keys for scATACseq

danlu1:
Add new variables used in PEC ATACseq template.


pitviper6:
@danlu1  Is adding the sampleBarcode key a different pull request since it's already in the schema?

PR #916, opened 2022-05-26
miscellaneous AD metadata updates

avanlinden:
Adding several values to existing terms and creating one new term for multiome data.

### DIFF
--- a/terms/ngs/libraryPrep.json
+++ b/terms/ngs/libraryPrep.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/sage.annotations-ngs.libraryPrep-0.0.10",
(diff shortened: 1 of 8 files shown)


terms/ngs/libraryPrep.json
```diff
@@ -44,6 +44,11 @@
             "description": "RNA library with size selection for microRNAs",
             "source": "Sage Bionetworks"
         },
+	{
+            "const": "mulitome",
```
tmzintel, line 48: there is a typo here - 'mulitome'

tmzintel (changes requested):
I left a comment on the suggested `libraryPrep.json` changes - there's a typo that should be fixed before accepting changes

terms/ngs/libraryPrep.json
avanlinden: Oh my goodness, good catch!

tmzintel (approved):
Looks good!

PR #922, opened 2022-07-12
add mat and nirs fileFormat for 1kD project

danlu1:


calkinsh (approved):
these look good to me

PR #931, opened 2022-10-04
add tab as a new fileFormat

danlu1:


pitviper6 (approved):

danlu1:
> @danlu1 Is this different from `tsv` (already in the dictionary)?

Yup but with different file extension.

avanlinden:
@danlu1 Ok thanks for clarifying!

PR #937, opened 2022-11-09
Update r-lib/setup-r branch versions in workflows

avanlinden:


danlu1 (approved):

danlu1:
Attached a reference here: https://github.com/r-lib/actions

PR #976, opened 2023-07-07
update R version for GH action and renv.lock

danlu1:


MasterPastures (approved):
Everything looks good!

PR #989, opened 2023-08-28
Update tissue.json

MasterPastures:
Added medial entorhinal cortex, lateral entorhinal cortex


avanlinden (changes requested):
Can you change the source values to be just the permanent URL? e.g. "http://purl.obolibrary.org/obo/UBERON_0007224". It's usually on the OLS term page. Other than that it looks good!

Screenshot 2023-08-28 at 1 16 11 PM

avanlinden (approved):
Looks good!

PR #992, opened 2025-03-28
Update macOS to v13 in register-schemas-dryrun.yaml

MasterPastures:
to try and avoid perpetual "Waiting for a runner to pick up this job..." message


MasterPastures:
not going to be using this repo

